PR #14, opened 2016-06-24
Update README.md + Fix repr.rs on containerDetails (change in docker remote API)

Morsicus:
Example in README couldn't compile. Compilation error : 

 `PullOptions` is the name of a struct or struct variant, but this expression uses it like a function name [E0423]
&PullOptions.builder().image("redis:2.8.18").build()


Morsicus:
I made only one PR for two differents things ... Sorry. If you prefer, i can make two differents PR :)

PR #15, opened 2016-07-01
add a container deletion api that takes options

gregwebs:
I have not actually tested this yet. However, it compiles with `cargo build` and I believe the interface is correct.

I added a new API `remove` so that `delete` retains backwards compatibility.


softprops:
Lloks good. Thanks man

PR #16, opened 2016-07-01
don't panic

gregwebs:
I don't know what the benefit of having a panic here is. It caught me off-guard when I make an API request for a container that was already destroyed.


softprops:
You're totally right.

PR #17, opened 2016-07-01
add a getter for the container id

gregwebs:


gregwebs:
It seems like this PR may have been missed among the group I sent. I am new to rust. Is there some other way I am supposed to get the id?

PR #26, opened 2016-12-26
Make id and status optional for events

r-darwish:
"id" and "status" are not present in network events. At the future I might create a pull request that adds support for different types of events, but at the moment this patch prevents the library from panicking when receiving a network event.


softprops:
Sounds good. In the near future I'll switching over to serde which may help make that easier

PR #27, opened 2016-12-26
Set the Host header

r-darwish:
It is currently impossible to create new container. This patch resolves this
issue. See the following link for details:

https://github.com/docker/docker/issues/20865


softprops:
Oh snap! Nice find and thanks for the patch. I'll try to get this in a release published in the next few days

softprops:
published a new version with this change included

PR #28, opened 2016-12-27
Implement the standard Error trait

r-darwish:
Implementing the standard Error trait allow composing the library error together
with frameworks such as error-chain


softprops:
@r-darwish thanks for the pull. Looks great!

softprops:
published a new version with this change included

PR #32, opened 2017-01-05
Add CapAdd to ContainerOptionsBuilder

blankenshipz:
Allows for creating `ContainerOptions` that extend additional privileges to the container.


softprops:
Thanks @blankenshipz

PR #33, opened 2017-01-06
OnBuild option is an array now.

shulcsm:


softprops:
published version 0.3.1 with this change

Issue #34, opened 2017-01-07
Add support for serializing hashes in ContainerOptions


blankenshipz:
Currently `ContainerOptions#serialize` can only deal with things that are `key`, `value` on the body of the message or `key, value` on `HostConfig` where each `HostConfig` key maps to an array of the values.

I think this falls down a bit when considering adding support for something like `Devices` where the value should be an array of key value pairs. 

The remote API (v 1.24) specifies devices like this: 

> Devices - A list of devices to add to the container specified as a JSON object in the form 

```JSON
{ "PathOnHost": "/dev/deviceName", "PathInContainer": "/dev/deviceName", "CgroupPermissions": "mrw"}

blankenshipz:
Closed by #36

PR #36, opened 2017-01-11
Add support for Devices

blankenshipz:
Attempting to address #34 

* adds a `params_hash` object to the `ContainerOptions` struct to track params that need to be serialized as arrays of objects
* adds a `devices` method to expose `params_hash` for adding "Devices"

In the current state creating a `device` would like like the following: 

```rust
    let mut opts = ContainerOptions::builder("alpine:latest");
    let mut devices = Vec::new();
    let mut device = HashMap::new();

    device.insert("PathOnHost".to_string(), "/dev/deviceName".to_string());
    device.insert("PathInContainer".to_string(), "/dev/deviceName".to_string());
    device.insert("CgroupPermissions".to_string(), "mrw".to_string());

    devices.push(device);

    opts.devices(devices);

    let container_options = opts.build();
```

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -398,6 +398,15 @@ impl ContainerOptionsBuilder {
         self
     }
 
+    pub fn devices(&mut self,
+                   devices: Vec<HashMap<String, String>>)
+                   -> &mut ContainerOptionsBuilder {
+        for d in devices {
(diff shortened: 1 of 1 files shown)


shulcsm:
Merge this please, i want to do labels so i need parms hash too.

softprops:
Sry for the delay. I'll try and pub a new version this weekend

softprops:
published version 0.3.1 with this change

PR #39, opened 2017-01-16
Fixed bug where `Images::build` panics on "status" replies

petehayes102:
I came across this bug when trying to build an image which relies on a repository that doesn't exist locally. The Docker API replies with "status" messages for downloading/extracting the base image.


softprops:
Nice catch. This seems to line up with thier latest docs https://docs.docker.com/engine/reference/api/docker_remote_api_v1.24/#create-an-image

PR #41, opened 2017-01-21
Store Cow<str> instead of &str

petehayes102:
Currently structures store a raw str slice, which makes for poor API ergonomics. By storing a `Cow<str>`, the struct can outlive the stack data if necessary.

I discovered the downside of the current API, which I queried here: https://users.rust-lang.org/t/binding-does-not-live-long-enough-even-though-its-being-returned/9002

Also, this is a good article on the topic, if you're interested: http://blog.jwilm.io/from-str-to-cow/

Btw, I also snuck in a commit to squash some warnings that were getting annoying!

### DIFF
--- a/examples/export.rs
+++ b/examples/export.rs
@@ -2,7 +2,6 @@ extern crate shiplift;
 
 use shiplift::Docker;
 use std::env;
-use std::io::prelude::*;
 use std::io::copy;
 use std::fs::OpenOptions;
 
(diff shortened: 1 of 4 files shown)


softprops:
Cool I've been using Into <String> interfaces in recent projects. I haven't really dug into Cow but this looks good. Thanks for the patch

petehayes102:
No dramas :) It's pretty new to me, too.

PR #56, opened 2017-04-20
Make rep-types clonable

pitkley:
While using this library I needed one of the types to derive from `Clone`, so I made all of them clonable.

I'm fairly new to Rust, so let me know if this is not desired.

### DIFF
--- a/src/rep.rs
+++ b/src/rep.rs
@@ -2,7 +2,7 @@
 
 use std::collections::HashMap;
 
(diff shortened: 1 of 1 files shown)


softprops:
It may only make binaries a little larger but I don't think it will hurt.

pitkley:
Cool, thanks for the quick response and the merge! 👍

PR #58, opened 2017-04-24
Add support for setting the log driver for a container

timmartin:
The log driver setting has a 3-level nested structure (`HostConfig.LogConfig.Type`) so I implemented a recursive function for mapping key names to JSON to replace the old special case for `HostConfig`.


softprops:
Thanks for the patch @timmartin!

Issue #65, opened 2017-09-06
Panic when building a new image


petehayes102:
When running `docker.images().build(...)`, it ends up panicking with this message:
>expected build output stream or error

This is because one of the JSON objects is not a "stream", "status" or "error":
>DEBUG:shiplift: Object({"aux": Object({"ID": String("sha256:3fdcaf0ab1b77ae77395563c779592fa6c00b210c38d5f83c6b01d5da2ccf2df")})})

Instead of identifying and repackaging each object, would it be better to simply return the JSON iterator? As far as I can tell, there isn't any value added by trying to repackage the objects into native structures. I also can't find a spec for the response JSON, so this issue might keep popping up over time. Manual doesn't help: https://docs.docker.com/engine/api/v1.31/#operation/ImageBuild

softprops:
I agree there have been a number of issues that have popped up like this in the past. I'll take note of this

petehayes102:
Cool. I solved it in my code by simply removing the map() fn, which also made the code a lot more readable. Would you accept a PR or do you want to consider other options?

softprops:
Absolutely open to prs. There's a longer term plan of upgrading to newer versions of hyper and serde which would allow for other things to happen and would definitely be breaking changes but if there are midterm solutions to current problems I accepts prs with open arms!

PR #66, opened 2017-09-07
Return RustcSerialize types instead of local types

petehayes102:
Closes https://github.com/softprops/shiplift/issues/65


softprops:
Travis error looks unrelated to the change. I'll follow up on that after

petehayes102:
Thanks mate! 😎

Issue #67, opened 2017-09-09
Filter on multiples events of the same type


jagu-sayan:
It's an awesome feature if we can listen on multiples events of the same type.

Like describe in [docker-cli events](https://docs.docker.com/engine/reference/commandline/events/) command.
```
Using the same filter multiple times will be handled as a OR;
for example --filter container=588a23dac085 --filter container=a8f7720b8c22
will display events for container 588a23dac085 OR container a8f7720b8c22
```

In this example, I want all container events that `start`, `stop` or `die`. But I only get  `die` event.
[Look documentation](http://lessis.me/shiplift/shiplift/builder/enum.EventFilter.html)
```rust
use shiplift::builder::EventsOptions;
use shiplift::builder::EventFilterType;
use shiplift::builder::EventFilter;

let mut options = EventsOptions::builder();
options.filter(vec![
    EventFilter::Type(EventFilterType::Container),
    EventFilter::Event("start".to_string()),
]);
options.filter(vec![
    EventFilter::Event("stop".to_string()),
]);
options.filter(vec![
   EventFilter::Event("die".to_string()),
]);
```

This is due to [internal implementation](https://github.com/softprops/shiplift/blob/1d95f805f1287aac532d04e7d3b7351d72f16950/src/builder.rs#L685) that only send one `filters` parameter to docker daemon.

jagu-sayan:
With this fix, you can now do
```rust
use shiplift::builder::EventsOptions;
use shiplift::builder::EventFilterType;
use shiplift::builder::EventFilter;

let mut options = EventsOptions::builder();
options.filter(vec![
    EventFilter::Type(EventFilterType::Container),
    EventFilter::Event("start".to_string()),
    EventFilter::Event("stop".to_string()),
    EventFilter::Event("die".to_string()),
]);
```
and get `start`, `stop` and `die` event of containers (and not only `die` event)

jagu-sayan:
@softprops ping

Nurrl:
Is this issue can be closed ? It seems to be merged :D

jagu-sayan:
@Nurrl  Yes :)

PR #68, opened 2017-09-10
Fix: Filter on multiples events of the same type

jagu-sayan:
Related to https://github.com/softprops/shiplift/issues/67

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,6 +22,8 @@ openssl = "0.9"
 rustc-serialize = "0.3"
 tar = "0.3"
 url = "0.5"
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -22,6 +22,8 @@ openssl = "0.9"
 rustc-serialize = "0.3"
 tar = "0.3"
 url = "0.5"
+serde = "1.0"
```
softprops: +1

softprops:
travis seems to be failing but for what seems unrelated problem. I'll deal with that

PR #72, opened 2018-04-01
[WIP] Refactor

matthiasbeyer:
This is my work-in-progress effort of rewriting this crate.

TODOs:

- [x] introduce proper error framework: error_chain
- [ ] Refactor codebase to be readable
- [ ] Make crate crash-free
    - [ ] Remove calls to `unwrap()` and `expect()`
    - [ ] Remove calls to `unreachable!()` and `unimplemented()`
- [x] Switch to serde as serialization framework

There's possibly more to do, I'm adding more todos in the process.

I'm too tired to continue today... maybe tomorrow.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,6 +24,7 @@ tar = "0.3"
 url = "0.5"
 serde = "1.0"
 serde_derive = "1.0"
+error-chain = "0.11"
 
 [dev-dependencies]
 env_logger = "0.4.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,6 +24,7 @@ tar = "0.3"
 url = "0.5"
 serde = "1.0"
 serde_derive = "1.0"
+serde_json   = "1.0"
 error-chain = "0.11"
 
 [dev-dependencies]
--- a/examples/containercreate.rs
+++ b/examples/containercreate.rs
@@ -4,7 +4,7 @@ use shiplift::{ContainerOptions, Docker};
 use std::env;
 
 fn main() {
-    let docker = Docker::new();
+    let docker = Docker::new().unwrap();
     let containers = docker.containers();
     if let Some(image) = env::args().nth(1) {
         let info = containers
--- a/examples/containerexec.rs
+++ b/examples/containerexec.rs
@@ -4,7 +4,7 @@ use shiplift::{Docker, ExecContainerOptions};
 use std::env;
 
 fn main() {
-    let docker = Docker::new();
+    let docker = Docker::new().unwrap();
     let options = ExecContainerOptions::builder()
         .cmd(vec![
             "bash",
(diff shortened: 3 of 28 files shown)


softprops:
@matthiasbeyer this is looking awesome ( and making me feel bad I haven't had time to put more work into this lately ). Would you like to be added as contributor to give you more commit freedom on this? I could use the help!

matthiasbeyer:
I'd love to be a contributor, sure.

Note, though, that from 13th of May until early 2019 I will be on vacation (yes, 2019!) so I won't be able to do anything at all during that time. Thus, best would be if you leave me at regular contributor, without special rights for the repository, and if I continue working on this in 2019 I would be honored to become a special-rights-contributor then!

Your choice! :heart:

matthiasbeyer:
@softprops I need your help here now.

The build succeeds on my local machine, but the tests fail because the schema how the objects are serialized has changed.

For example, in the `container_options_simple` test, the following is expected

```json
{"HostConfig":{},"Image":"test_image"}
```

But the new schema looks like this:

```json
{"name":null,"params":{"Image":"test_image"},"params_hash":{},"params_list":{}}
```

That is because serde serializes in another way than rustc_serialize did and because I did not adapt the format of the structs yet. I'm not entirely sure for what this JSON is used. Could you explain?

If the JSON is used for communicating with docker, I guess I need to adapt the code so that it fits the tests, but if not I'd rewrite the tests.

So please tell me: What's the _purpose_ of this JSON. :smile:

softprops:
@matthiasbeyer sent contributor invite.

softprops:
the docs for docker's rest api are much better than they used to be but it looks like the container create docs can be found [here](https://docs.docker.com/engine/api/v1.36/#operation/ContainerCreate) the rust struct for the options are used [here](https://github.com/softprops/shiplift/blob/2cdb3723d2a976c8e4260bfd37276ad32c20d88e/src/lib.rs#L503)

matthiasbeyer:
First: Thanks for the contributor rights.

I tried to generate a API from the openAPI specifications docker provides, but the resulting code does not compile, unfortunately. I filed an issue for this in https://github.com/swagger-api/swagger-codegen/issues/7964

Lets see what happens. If we can resolve this issue, I'd like to merge the `shiplift-swagger` repository into this repository and rebuild shiplift using the generated code. There's probably little to do then, which I'd welcome. Not sure though.

If this does not work, we can continue develop this PR here. But it would be tedious work given that we could generate it all.

sharksforarms:
Hey @matthiasbeyer @softprops 

I could dedicate some time to re-factor, working on this PR atm https://github.com/softprops/shiplift/pull/74

softprops:
@matthiasbeyer there've been some recent changes that seem to have introduced a merge conflict here. I'm thinking it may be a good idea to close this since it was side in scope in favor of smaller, more focused pulls. What do you think?

matthiasbeyer:
Yes, I guess closing would be a good idea! Awesome that work is happening here!

PR #91, opened 2018-09-24
Update to hyper 0.12

abusch:
This builds upon @dylanmckay's branch that did most of the work. I've bumped hyperlocal to 0.6, removed the dependency on `tokio_core`, and now store an instance of `Runtime` in `Transport` instead of creating a new one for every request. I'm not sure which of these changes did it, but it fixes the hangs I was having with this branch.


src/lib.rs
```diff
@@ -628,75 +628,97 @@ impl Docker {
...
+        match env::var("DOCKER_HOST").ok() {
+            Some(host) => {
+                let host = host.parse().expect("invalid url");
+                Docker::host(host)
+            },
+            None => Docker::unix("/var/run/docker.sock")
+        }
+    }
+
+    /// Creates a new docker instance for a dockr host
```
dylanmckay: `s/dockr/docker`, it's potentially my typo originally though

src/transport.rs
```diff
@@ -2,56 +2,71 @@
...
 /// with the docker daemon
 pub enum Transport {
     /// A network tcp interface
-    Tcp { client: Client, host: String },
+    Tcp {
+        client: Client<HttpConnector>,
+        runtime: RefCell<Runtime>,
+        host: String,
+    },
+    /// TCP/TLS.w
```
dylanmckay: trailing `w`

src/transport.rs
```diff
@@ -148,23 +187,48 @@ impl Transport {
...
-            .and_then(|x| x.as_object())
-            .and_then(|x| x.get("message"))
-            .and_then(|x| x.as_string())
-            .map(|x| x.to_owned());
-
-        message
-    } else {
-        None
+fn get_error_message(res: Response<Body>) -> Option<String> {
+    let chunk = match res.into_body().concat2().wait() {
```
dylanmckay: Most of the other `wait` calls have been replaced with `self.runtime().block_on(req)` - should this one be too?

src/transport.rs
```diff
@@ -2,56 +2,71 @@
...
-        mime::TopLevel::Application,
-        mime::SubLevel::Ext(String::from("tar")),
-        vec![],
-    ))
+use std::cell::{RefMut, RefCell};
+use std::io::{BufReader, Cursor};
+use std::io::{Read, Write};
+use tokio::runtime::Runtime;
+
+trait InteractiveStream : Read + Write { }
```
dylanmckay: I suspect this might be left over from when I first split this PR out from some related shiplift changes, it doesn't look used.

src/transport.rs
abusch: hmm, I guess so?

abusch:
Thanks for the review! I think I've addressed all the comments.

dylanmckay:
Perfect, those four comments have all been fixed

abusch:
@softprops what say you? Do you think this can be merge?

softprops:
It's been a while. I'm going to try to fold as much as I can into a new release including this

softprops:
Can you fix up the remaining merge conflicts. I'll merge after and try to roll up a release

abusch:
Alright, I've merged master in this branch and resolved the conflicts. It builds and the tests pass both with and without the feature.
(not to self: I should put SSL support behind a feature flag as well...)

src/transport.rs
```diff
@@ -63,112 +81,150 @@ impl Transport {
...
+    {
+        let req = self.build_request(method, endpoint, body)?;
+        let res = self.send_request(req)?;
+
+        match res.status() {
+            StatusCode::OK
+            | StatusCode::CREATED
+            | StatusCode::SWITCHING_PROTOCOLS => {
+                let chunk =
+                    self.runtime().block_on(res.into_body().concat2())?;
```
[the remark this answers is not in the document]
abusch: Yes, I suspect you're correct... The whole wrapping of Hyper's async API in a synchronous API is quite clunky unfortunately :( Ideally this would return a `Stream<Vec<u8>>` or similar, and I'm slowly working towards this in a branch but I'm not there yet... I'm not sure how to fix this until then, so I'm open to ideas :)

PR #92, opened 2018-09-27
Make Unix socket support optional

fkrull:
And by making that optional, the hyperlocal dependency can be made optional as well (= hidden behind the "unix-socket" feature). And without the hyperlocal dependency, the project can be used on Windows; obviously only over HTTP(S).

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -17,6 +17,7 @@
 extern crate log;
 extern crate hyper;
 extern crate hyper_openssl;
+#[cfg(feature = "unix-socket")]
 extern crate hyperlocal;
 extern crate flate2;
 extern crate jed;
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -641,14 +641,19 @@ impl Docker {
...
                     transport: Transport::Unix {
-                        client: Client::with_connector(UnixSocketConnector),
+                        client: Client::with_connector(hyperlocal::UnixSocketConnector),
                         path: host.path().to_owned(),
                     },
                 }
             }
+
+            #[cfg(not(feature = "unix-socket"))]
+            "unix" => panic!("Unix socket support is disabled"),
```
softprops: Can we add a note in the error message about how on could to enable it? The path to discovery is not obvious to those that are new to rust and cargo.

fkrull: i'm not sure that's necessary? It's on by default, so to see this error, one would have to first go and explicitly disable it.

PR #110, opened 2018-10-05
Update CONTRIBUTING.md

ahmetcetin:
added missing link of issues list, which was deleted by mistake in previous commit

## What did you implement:

Closes: #xxx


softprops:
thank you :)

Issue #64, opened 2017-08-03
See over lifetime parameter and borrowing API


bluss:
Hi, sorry for the drive through comment; noticed this while answering questions from users of your library. I hope this helps.

In shiplift 0.3.2 I see this function:

```rust
pub fn get(&'a self, name: &'a str) -> Container
```

using `'a` like this -- it's not a parameter of the current method, but of the whole impl block -- is almost never what you want. The result is that the code is asking for far more strict borrowing rules than are actually required.

Remember: **If a lifetime *parameter* is used in two places, those two borrows must have the same *lifetime.***

- `&'a self` says that the borrow of self *just to call this method* must be the same length as the whole borrow of the reference that `Containers` is already storing.
  - What you actually want: The borrow of self can be shorter too, that's fine and applies in many more situations

- The `&'a str` says that the borrow of the string, too, must be of the same scope and lifetime. It adds another variable into the same borrow, which just makes it even more complicated to find a situation where they all live that same long time.

What you probably want:

```rust
    // I'm giving the lifetimes unconventional names here
    // so that we see them differently
    pub fn get<'thestring>(&self, name: &'thestring str) -> Container<'a, 'thestring> {
        Container::new(self.docker, name)
    }
```

What happened here? The `&'a Docker` is just a *shared* reference (im/mutability is important), so we just copy it along. Don't involve `self` in this borrow. Just read & copy the reference to the output.

The string gets its own lifetime variable. Now the signature describes the most general (and most versatile) borrowing behavior that is actually happening.

I went into the details for this method, and I hope it helps with fixing a few issues here and there in the API.

PR #113, opened 2018-10-08
Fix some lifetime issues

abusch:
## What did you implement:

A few methods used some lifetimes that were too restrictive, so relax them.

Closes: #64

The tests still pass


softprops:
@abusch I just fixed up rustfmt after the big serde change. I think that introduced a merge conflict here. Mind pulling the latest master into this branch?

softprops:
thank you sir :)

Issue #123, opened 2018-10-13
Library broken by recent changes.


Nurrl:
- Crate version: master
- OS: Linux x64
- Output of running `docker version` on the command line:
```
Client:
 Version:           18.06.1-ce
 API version:       1.38
 Go version:        go1.10.3
 Git commit:        e68fc7a
 Built:             Sun, 16 Sep 2018 13:21:33 +1000
 OS/Arch:           linux/amd64
 Experimental:      false

Server:
 Engine:
  Version:          18.06.1-ce
  API version:      1.38 (minimum version 1.12)
  Go version:       go1.10.3
  Git commit:       e68fc7a
  Built:            Sun Sep 16 03:21:33 2018
  OS/Arch:          linux/amd64
  Experimental:     false
```

So as in the title, the recent changes broke the `Images::pull()` which results in a: 
```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: SerdeJsonError(Error("invalid type: map, expected a sequence", line: 1, column: 1))', libcore/result.rs:945:5
```
for no reason, but the download occurs, it successfully pulls the image.
And after that, `Containers::create()` fails with `Config cannot be empty in order to create a container` when the config is actually not empty:
```rust
&ContainerOptionsBuilder::new("redis:latest")
        .name("test-name").build()
```

All that seems really weird to me..

Nurrl:
Fixed the second issue, sending PR asap.

softprops:
Note there a number of changes recently folded into master the upcoming release https://github.com/softprops/shiplift/blob/master/CHANGELOG.md. sorry if this caught you off gaurd. And pulls are def welcome!

Nurrl:
@FauxFaux, That's what I was noticing, I'm gonna try it out !

FauxFaux:
I have it working, give me a minute!

Nurrl:
Oh nice !

FauxFaux:
Mmm. This new API is not great for `build()`, as it now only returns results at the end of the build, not live progress.

Nurrl:
We said the same but in different places :,)

Nurrl:
Let's continue the talk here, #126 should be merged for now since it fixes the issue temporarily.

Nurrl:
But #126 should't close this issue anymore. Because it's not a perfect integration.

PR #124, opened 2018-10-13
Fixed error causing bad json to be sent to Docker API.

Nurrl:
## What did you implement:

Fixes the 2nd part of #123 

Finding that 2nd part of #123 isn't broken now.

Nothing, just a hot-patch.


Nurrl:
Didn't used `cargo fmt` because I prefer having rust with `apt` and not with `rustup`.

PR #126, opened 2018-10-14
serde can't read a Vec<Value> from a stream

FauxFaux:
## What did you implement:

Use the `serde_json` api correctly for streams of data. (example in ticket)

Closes: #123

Works for my app: https://github.com/FauxFaux/fappa/commit/5ff21c38ef555f0ff5f5f0134fef6c8c175cf3d1#diff-639fbc4ef05b315af92b4d836c31b023R138

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -180,7 +180,12 @@ impl<'a> Images<'a> {
 
         self.docker
             .stream_post(&path.join("?"), Some((Body::from(bytes), tar())))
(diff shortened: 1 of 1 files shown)


Nurrl:
It works for my project too, but note that it's not real-time anymore.

Nurrl:
Now it don't work anymore again .-. That's really weird

softprops:
@Nurrl thanks for the fix. Please be careful with the language you use in these threads.

Issue #73, opened 2018-04-12
Exposed ports


sharksforarms:
Hello, is it possible to add ports to expose when creating a docker container?

Nurrl:
I don't think so :c @sharksforarms I'll implement this ASAP, hopefully they will merge :)

PR #74, opened 2018-04-13
[WIP] Expose port

sharksforarms:

PR #127, opened 2018-10-15
[OK] Expose port

Nurrl:
## What did you implement:

```rust
ContainerOptionsBuilder::expose();
```
It adds the ability to expose ports from Containers.

Closes: #73 
Replaces: #74 

By testing and with units tests.

Nothing.


Nurrl:
Arrrgh, I can't use `cargo fmt` :c

softprops:
thanks @Nurrl!

Issue #111, opened 2018-10-08
Make the API asynchronous


abusch:
Since we're using hyper under the hood, which is fully async, it is a bit of a shame that shiplift's API isn't. It would be great to change the methods that make a call to Docker to return a `Future` of some kind.

Some examples of Rust crates with an Async API:
* [fantoccini](https://github.com/jonhoo/fantoccini)
* [tokio-zookeeper](https://github.com/jonhoo/tokio-zookeeper)

Maybe for version 0.5? :)

softprops:
I agree. This should be pretty straight forward There are many calls to [block_on](https://github.com/softprops/shiplift/blob/master/src/transport.rs#L195) that seem unnecessary.

abusch:
I've been working on it and am slowly making progress. Still a few methods to do, and some clean-up, but the `stats` example works and properly streams `Stats` objects as they're received, which I'm quite happy about :) You can see the work in progress here: https://github.com/abusch/shiplift/tree/async_api

PR #128, opened 2018-10-16
Async api

abusch:
## What did you implement:
Migrate the whole API to an asynchronous style i.e. every method returns either a `Future` or a `Stream` which will need to be executed in a Tokio runtime by the user. This remove any use of `block_on()` internally, as well as the need to hold a `Runtime` inside `Transport`. This allows proper streaming (e.g. for stats or events), and integration with other async crates.

Closes: #111 

I haven't really yet, except by running some of the examples :)

This is a major API change that affects pretty much every method.

## Things that still need to be done:
- [x] Finish implementing the methods I've commented out (see `TODO(abusch)` in the code)
- [x] Finish porting the remaining examples
- [ ] Introduce a `DockerFuture<T>` type (and `DockerStream<T>`?) to simplify method signature (maybe?)
- [x] Fix tests and doc tests
- [x] Clean-up imports / clippy warnings


examples/events.rs
```diff
@@ -1,11 +1,19 @@
 extern crate shiplift;
+extern crate tokio;
 
 use shiplift::Docker;
+use tokio::prelude::*;
```
softprops: When possible I try to avoid wildcards imports because the hide potentially useful documentation. Since these examples files serve as documentation I'd like it to be a bit more clear. What prelude items does this file use/need?

src/lib.rs
```diff
@@ -774,76 +712,106 @@ impl Docker {
...
+    pub fn ping(&self) -> impl Future<Item = String, Error = Error> {
         self.get("/_ping")
     }
 
     /// Returns an interator over streamed docker events
-    pub fn events(
-        &self,
-        opts: &EventsOptions,
-    ) -> Result<Vec<Event>> {
+    pub fn events(&self, opts: &EventsOptions) -> impl Stream<Item = Event, Error = Error> {
```
softprops: 😍 good choice. streams are an awesome choice. Doesn't have to happen here but I just noticed the pre-existing typo in this rustdoc: `interator` -> `iterator`. it was both incorrect and a typo. we may just want to say something like `Returns a stream of docker events`. The docs for this crate could use some polish but that can definately be a followup

src/transport.rs
```diff
@@ -71,15 +63,14 @@ impl Transport {
...
     {
-        let mut res = self.stream(method, endpoint, body)?;
-        let mut body = String::new();
-        res.read_to_string(&mut body)?;
-        debug!("{} raw response: {}", endpoint, body);
-        Ok(body)
+        self.stream(method, endpoint, body)
+            .concat2()
+            .and_then(|v| String::from_utf8(v).map_err(Error::Encoding).into_future())
+        // debug!("{} raw response: {}", endpoint, body);
```
softprops: What do you think about keeping the `debug!` log? I've often found these useful in cases where I need to reproduce a deserialization problem with a type mapping. The errors that bubble up from serde sometimes do not contain enough contextual information to give you an idea of what is in error.

src/transport.rs
```diff
@@ -121,107 +112,68 @@ impl Transport {
...
-        }
+        self.send_request(req)
+            .and_then(|res| {
+                let status = res.status();
+                match status {
+                    StatusCode::OK
+                    | StatusCode::CREATED
+                    | StatusCode::SWITCHING_PROTOCOLS
+                    | StatusCode::NO_CONTENT => future::ok(res),
+                    _ => future::err(Error::Fault {
```
softprops: much much much nicer!

src/transport.rs
```diff
@@ -121,107 +112,68 @@ impl Transport {
...
-                    .and_then(|x| x.as_str())
-                    .map(|x| x.to_owned())
-            }
-            Err(..) => None,
-        }
-    }
+    // Extract the error message content from an HTTP response that
+    // contains a Docker JSON error structure.
+    // fn get_error_message(&self, body: &str) -> Option<String> {
+    //     match String::from_utf8(chunk.into_iter().collect()) {
```
softprops: now that we're using serde, if there's a standard way to represent it with a struct. we have more civilized tools now!

softprops:
@abusch I just merged a pull for exposing ports on the container builder which caused a conflict with this branch. Just wanted to give you a heads up. I took a look over this branch. It looks awesome! Thanks for taking all of this on.

examples/events.rs
abusch: Fair enough. I thought this was the recommended way to use Tokio.

src/transport.rs
abusch: Yeah, I'm struggling a bit as to how best to reintroduce this method, because in the error case we need to read the whole body and parse it to get the error message, but in the success page we just want to return a stream of body chunks... Haven't find the right sequence of combinators yet :)

abusch:
Only `Container::exec()` left, which is... a bit tricky.

abusch:
@softprops I think this might be ready for another round of review, whenever you get a chance. I'm kind of torn about introducing a `DockerFuture<T>` type... It might simplify the signature a bit, and remove some usages of `Either` in the code, but not sure if it's worth the trouble. What do you think?

Also pinging @srijs in case you have suggestions/comments on my Hyper/Tokio/Future code :)

softprops:
Crap there's one tiny merge conflict that crept in. Can you merge master to resolve real quick and I'll merge. Thanks for your patience on this

abusch:
No worries. I'll resolve the conflict shortly on my commute home :)

abusch:
All checks are green!

softprops:
Awesome work. I'm still on vacation but I'll try to get a release out soon. I kind of gave up on release notes for this one. I'm going to try doing releases more often for this repo to ease that burden

schrieveslaach:
@abusch, I tried to use the new async API and I'm wondering how is it possible to collect data. For example, I want to load some information from some containers and provide a REST response.  Could you provide any examples?

danieleades:
i've been using shiplift with Rocket to create a web interface for managing docker containers. I'm not even sure how to begin using this new API with Rocket. I'm not even sure that I can. I think this is probably a breaking (or blocking) change on any project where the threading model is managed by a library that doesn't use tokio. I get that async in rust is shiny and exciting right now, but does this implementation not make your library far less composable?

schrieveslaach:
@danieleades, I managed to use the new API behind a REST API with rocket. Have a look at [here](https://github.com/aixigo/PREvant/blob/master/api/).

danieleades:
@schrieveslaach, thanks for taking the time. Is there any overhead for spooling up a new runtime instance every time a request is handled? How will that scale?

schrieveslaach:
@danieleades, I do not have any numbers and I think that the current implementation of PREvant is slower with the creation of a runtime for each docker command. I need to address this issue when I'm back at work.

PR #130, opened 2018-10-25
Make TLS/openssl an optional dependency

rsdy:
## What did you implement:

OpenSSL should be an optional dependency, because a hard dependency makes building a static binary using shiplift impossible.

```
% cargo build --no-default-features --features unix-socket &>/dev/null && echo $?
0
% cargo build --no-default-features --all-features &>/dev/null && echo $?
0
```

The dependency on OpenSSL is now optional, and can be controlled using the `tls` feature.


softprops:
sry for the delay @rsdy I've been on a long vacation. i just rolled through a bunch of open prs leaving this one with some merge conflicts. would you mind merging master. pull merge after. thanks!

softprops:
heads up, im going to try and upgrade to 2018 edition before the next release here https://github.com/softprops/shiplift/pull/141.

rsdy:
I'm happy to pick up on this patch in a few days, and will follow up with the changes.

rsdy:
Sorry, I followed up on the changes then forgot to ping back. The tests pass now on this branch, but please take another look just to be on the safe side.

softprops:

PR #134, opened 2018-12-07
Support for setting CPU shares/memory for image builder

dylanmckay:
This adds support for setting CPU shares and the memory limit when creating images and containers.

I used a the patch in a project so that it wouldn't take so long to rebuild an image I often rebuild.

* Support for setting allocation of CPU shares memory when creating images and containers has been added.


softprops:
@dylanmckay you're killing it with this pulls

PR #135, opened 2018-12-07
Remove an unused type parameter from the 'nocache' function

dylanmckay:
This removes an unused type parameter from the 'nocache' function

The function is currently unusable because the type parameter has no
uses or bounds.

Callers of the function must explicitly specify the type `R`, even though the specific type is meaningless.

I could successfully call `nocache` without type explicit type parameters with the patch.


softprops:
@dylanmckay 👍

PR #136, opened 2018-12-07
Support interactive stdin/stdout streams

dylanmckay:
This adds support for streaming stdin, stderr, and stdout independently
to a running container.

The underlying API is futures-based, meaning the code is implemented
asynchronously. A synchronous API is also exposed, which is implemented
by simply waiting on the asynchronous API futures.

This also modifies the existing Tty logic so that the storage type of
the data is a Vec<u8> rather than a String. This is also how the Rust
standard library persists data from the standard streams. In my
particular application, I'm using stdin/stdout as the communication
method between a container a host application. In it, a byte-based protocol is
used.

Streaming works by performing a TCP upgrade; upgrading a higher-level
HTTP connection to a lower-level TCP byte stream upon agreement with the
server. Docker will automatically upgrade HTTP container log requests to
TCP byte streams of a custom std{in,out,err} multiplexing protocol if
the client requests it with the 'Connection: Upgrade' header.

I've got a project that implements a nontrivial byte-based protocol to send command/reply frames from the host app to the container over stdout/stdin. I can successfully parse, send, and process frames in both directions asynchronously with this patch.

  * The `Container::logs` function now returns a `tty::Chunk` type containing raw bytes, rather than a `TtyLine` object containing a UTF-8 string.
  * The standard streams of container logs can now be independently managed and communicated with

### DIFF
--- a/examples/containerexec.rs
+++ b/examples/containerexec.rs
@@ -25,10 +25,11 @@ fn main() {
         .containers()
         .get(&id)
         .exec(&options)
-        .for_each(|line| {
-            match line.stream_type {
-                StreamType::StdOut => println!("Stdout: {}", line.data),
-                StreamType::StdErr => eprintln!("Stderr: {}", line.data),
+        .for_each(|chunk| {
+            match chunk.stream_type {
+                StreamType::StdOut => println!("Stdout: {}", chunk.as_string_lossy()),
+                StreamType::StdErr => eprintln!("Stderr: {}", chunk.as_string_lossy()),
+                StreamType::StdIn => unreachable!(),
             }
             Ok(())
         })
--- a/examples/logs.rs
+++ b/examples/logs.rs
@@ -14,10 +14,11 @@ fn main() {
         .containers()
         .get(&id)
         .logs(&LogsOptions::builder().stdout(true).stderr(true).build())
-        .for_each(|line| {
-            match line.stream_type {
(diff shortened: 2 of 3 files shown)


src/transport.rs
```diff
@@ -182,6 +185,54 @@ impl Transport {
...
+        };
+
+        let req = self.build_request(method, endpoint, body, |builder| {
+            builder.header(header::CONNECTION, "Upgrade")
+                .header(header::UPGRADE, "tcp");
+        }).expect("Failed to build request!");
+
+        self.send_request(req).and_then(|res| {
+            if res.status() != StatusCode::SWITCHING_PROTOCOLS {
+                panic!("Our server didn't upgrade: {}", res.status());
```
abusch: It might be better to return a failed Future here, instead of panicking?

dylanmckay: Good idea

dylanmckay: Fixed

dylanmckay:
I've also added `Chunk::as_string`, `Chunk::as_string_lossy` accessors, and updated the examples to use them.

Tests pass on my machine, I can compile all examples, and all tests pass.

One of the existing PRs to fix the existing `rustfmt` error in `shiplift/master` need to be merged before CI will pass for this change. I've included it in one of my smaller PRs, #134, but it looks like #137 also does it.

PR #137, opened 2018-12-14
Support for AutoRemove flag

sjanota:
## What did you implement:

I've added support for `HostConfig.AutoRemove` parameter when creating new container.

1. I modified one of the unit tests to include this parameter.
2. I've tested if container is properly removed after execution using this snippet: 
```rust
use shiplift::{ContainerOptions, Docker};
use tokio::prelude::Future;

fn main() {
    let docker = Docker::new();
    let create_opts = ContainerOptions::builder("alpine:3.7")
        .auto_remove(true)
        .name("test")
        .cmd(vec!["echo", "123"])
        .build();
    let create_fut = docker
        .containers()
        .create(&create_opts)
        .map(move |info| {
            println!("{:?}", info);
            let start_fut = docker
                .containers()
                .get(info.id.as_ref()).start()
                .map_err(|e| eprintln!("Error: {}", e));
            tokio::spawn(start_fut);
        })
        .map_err(|e| eprintln!("Error: {}", e));
    tokio::run(create_fut);
}
```

Added support for `--rm`-like behaviour.

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -578,6 +578,15 @@ impl ContainerOptionsBuilder {
         self
     }
 
+    pub fn auto_remove(
+        &mut self,
+        set: bool
+    ) -> &mut Self {
+        self.params
+            .insert("HostConfig.AutoRemove", json!(set));
+        self
+    }
+
     pub fn build(&self) -> ContainerOptions {
         ContainerOptions {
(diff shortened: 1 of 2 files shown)


softprops:
@sjanota thank you sir!

PR #149, opened 2019-01-07
Add ContainerOptionsBuilder::privileged()

### LINKED ISSUE #148: Add support for creating privileged docker containers
Essentially it is a `docker run --privileged ...`. Looks like it is impossible right now.

imp:
This adds new `ContainerOptionsBuilder` method `privileged()` to enable running privileged docker containers. I think it should be mentioned in CHANGELOG.md. Also, this could be potentially an incompatible change, since API addition is considered an incompatible change in semver universe.
 
The implementation is pretty straightforward, - the `privileged` field is already present in `HostConfig`, so there is only a new builder option to set the field to a required value.

As a side note - I have followed the precedence set by `auto_remove()` that gets a boolean argument. However, if I was to implement it from the scratch I'd choose to drop the boolean argument completely (similarly to the `docker` command line - you only mention it if you need it).

Fixes #148

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -661,6 +661,14 @@ impl ContainerOptionsBuilder {
         self
     }
 
+    pub fn privileged(
+        &mut self,
+        set: bool,
+    ) -> &mut Self {
+        self.params.insert("HostConfig.Privileged", json!(set));
+        self
+    }
+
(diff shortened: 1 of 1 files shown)


softprops:
@imp this branch seems to have a conflict with master. if you pull merge in latest Ill merge here

softprops:

PR #150, opened 2019-02-12
Copy from container

bossmc:
## What did you implement:

Added `copy_from` to `Container` - this returns a `Stream` of `Vec<u8>`s (similar to `export`) the stream is actually a `Tarball` of the requested files (as you can see in the example).

I've also run `clippy`/`fmt`/`upgrade` over the codebase.

* Added `example/containercopyfrom`

* Nothing beyond the new function.


softprops:

PR #152, opened 2019-02-22
Migrate serde dependency to use derive feature

veeg:
This is in line with best practices recommended by serde[[1]].
This will also resolve downstream crates depending on shiplift who enable the
serde derive feature, due to Cargos unification of features for each crate[[2]].

[1]: https://github.com/serde-rs/serde/issues/1441
[2]: https://github.com/rust-lang/cargo/issues/4361#issuecomment-348538243

Example of the problem for downstream crates:

```
$ cargo new deptest
$ cd deptest
$ echo 'shiplift = "0.4"' >> Cargo.toml
$ echo 'serde = { version = "1.0", features = ["derive"] }' >> Cargo.toml
$ cargo build
```

This will result in the following error:
```
error[E0252]: the name `Serialize` is defined multiple times
 --> /home/veeg/.cargo/registry/src/github.com-1ecc6299db9ec823/shiplift-0.4.0/src/builder.rs:5:5
  |
4 | use serde::Serialize;
  |     ---------------- previous import of the macro `Serialize` here
5 | use serde_derive::Serialize;
  |     ^^^^^^^^^^^^^^^^^^^^^^^ `Serialize` reimported here
  |
  = note: `Serialize` must be defined only once in the macro namespace of this module
help: you can use `as` to change the binding name of the import
  |
5 | use serde_derive::Serialize as OtherSerialize;
  |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

error: aborting due to previous error
```

## What did you implement:

Closes: #xxx

`cargo build && cargo test`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,8 +34,7 @@ tokio = "0.1.15"
 tokio-codec = "0.1.1"
 tokio-io = "0.1.11"
 url = "1.7.2"
-serde = "1.0.87"
-serde_derive = "1.0.87"
+serde = { version = "1.0.87", features = ["derive"] }
 serde_json = "1.0.38"
 
 [dev-dependencies]
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -2,7 +2,6 @@
 
 use crate::{errors::Error, Result};
 use serde::Serialize;
-use serde_derive::Serialize;
 use serde_json::{self, json, map::Map, Value};
 use std::{
     cmp::Eq,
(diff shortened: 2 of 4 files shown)


softprops:
Thanks for the patch. I've started noticing this pop up in other crates. Since serde is such a common crate it tends to have a ripple effect across the ecosystem :)

veeg:
I fell into this rabbit hole and had to figure out the cargo feature story. Hopefully they will make the huge refactoring work needed in cargo to split crate units by features enabled. I can image this issue with serde and how widespread it is will accelerate this process.

softprops:

Issue #143, opened 2018-12-24
docker.images.pull returns parsing error on `\r\n` line ending


softprops:
running `cargo run example imagepull busybox` yields an error 

`Error: Docker Error: trailing characters at line 2 column 1` after eventually receiving a `\r\n` line ending from the docker daemon.

- Crate version: lastest master
- OS: osx
- Output of running `docker version` on the command line:

```
docker version                                                         ✘ 1
Client: Docker Engine - Community
 Version:           18.09.0
 API version:       1.39
 Go version:        go1.10.4
 Git commit:        4d60db4
 Built:             Wed Nov  7 00:47:43 2018
 OS/Arch:           darwin/amd64
 Experimental:      false

Server: Docker Engine - Community
 Engine:
  Version:          18.09.0
  API version:      1.39 (minimum version 1.12)
  Go version:       go1.10.4
  Git commit:       4d60db4
  Built:            Wed Nov  7 00:55:00 2018
  OS/Arch:          linux/amd64
  Experimental:     true
```

softprops:
update. its related to the semantics of the `fromImage` parameter without a tag, the docker daemon will attempt to pull _all_ the tags. I think what was observed here was a delimiter between delimiters when switching from the first tagged image to the next

softprops:
I'll document the behavior for now but I don't think this is a release blocker

Qwaz:
I am experiencing the same issue with `hello-world:latest` even with the tag specified.

This is the code for reproducing the error. (`runtime` is [tokio current thread Runtime](https://docs.rs/tokio/0.1/tokio/runtime/current_thread/struct.Runtime.html))
```
runtime.block_on(env.docker.images().pull(&PullOptions::builder().image("hello-world").tag("latest").build())
    .then(|result| -> Result<Value, DockerError> {
        printer.write_line(&format!("{:?}", result));
        result
    }).collect());
```

It prints the below error in most of the case,
```
Ok(Object({"id": String("latest"), "status": String("Pulling from library/hello-world")}))
Err(SerdeJsonError(Error("trailing characters", line: 2, column: 1)))
```

... and occasionally prints the expected result.
```
Ok(Object({"id": String("latest"), "status": String("Pulling from library/hello-world")}))
Ok(Object({"status": String("Digest: sha256:b3a26e22bf55e4a5232b391281fc1673f18462b75cdc76aa103e6d3a2bce5e77")}))
[...]

PR #154, opened 2019-02-25
Support multiple messages per chunk in streaming image pull

aluminous:
## What did you implement:

When pulling images, sometimes multiple messages arrive in one chunk leading to errors about "trailing characters" (pointed out in https://github.com/softprops/shiplift/issues/143#issuecomment-450346853):

```
Error: Docker Error: trailing characters at line 2 column 1
```

This adds support for multiple messages per chunk, so that they are returned as if they arrived one at a time.

Using the `imagepull` example with Docker Engine 18.09.0 (Docker for Mac).

Before:
```
$ cargo run --example imagepull ubuntu:18.04
    Finished dev [unoptimized + debuginfo] target(s) in 0.33s
     Running `target/debug/examples/imagepull 'ubuntu:18.04'`
Object({"id": String("18.04"), "status": String("Pulling from library/ubuntu")})
Error: Docker Error: trailing characters at line 2 column 1
```

After:
```
$ cargo run --example imagepull ubuntu:18.04
   Compiling shiplift v0.4.0 (/Users/austin/src/shiplift)
    Finished dev [unoptimized + debuginfo] target(s) in 19.70s
     Running `target/debug/examples/imagepull 'ubuntu:18.04'`
Object({"id": String("18.04"), "status": String("Pulling from library/ubuntu")})
Object({"id": String("6cf436f81810"), "progressDetail": Object({}), "status": String("Pulling fs layer")})
[...lots of output...]
Object({"status": String("Status: Downloaded newer image for ubuntu:18.04")})
```

* Resolves an error due to "trailing characters" when pulling images.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -197,7 +197,15 @@ impl<'a> Images<'a> {
         self.docker
             .stream_post::<Body>(&path.join("?"), None)
             // todo: give this a proper enum type
-            .and_then(|r| serde_json::from_slice::<Value>(&r[..]).map_err(Error::from))
+            .map(|r| {
+                futures::stream::iter_result(
+                    serde_json::Deserializer::from_slice(&r[..])
+                        .into_iter::<Value>()
+                        .collect::<Vec<_>>(),
+                )
+                .map_err(Error::from)
+            })
+            .flatten()
     }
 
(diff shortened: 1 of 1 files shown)


softprops:

Issue #161, opened 2019-04-17
Support for publish?


gwihlidal:
There is an expose method on the container builder, but there is no way (it appears) to publish a port, which is an important feature - see: https://medium.freecodecamp.org/expose-vs-publish-docker-port-commands-explained-simply-434593dbc9a3

Is this a known issue? It would be awesome to get publish added to the options.

Thank you!

gwihlidal:
It looks like HostConfig.PortBindings is specified by expose(), but ExposedPorts is not?
https://stackoverflow.com/questions/20428302/binding-a-port-to-a-host-interface-using-the-rest-api

If I run my container from the cmd line:
```
docker run --detach --publish 63999:63999 --name svc-test-cmd my_image
```

vs from shiplift:
```
let fut = docker
        .containers()
        .create(&shiplift::ContainerOptions::builder("my_image")
        .name("svc-test-shiplift")
        .expose(63999, "", 63999)
        .build())
        .map(|info| println!("{:?}", info))
        .map_err(|e| eprintln!("Error: {}", e));
    tokio::run(fut);
```

You can see the lack of published ports with `docker ps -a`
```
PORTS                      NAMES
                           svc-test-shiplift
0.0.0.0:63999->63999/tcp   svc-test-cmd
```

softprops:
I would be open to a pull request to add this feature

PR #162, opened 2019-04-19
Added `ExposedPorts` mapping

gwihlidal:
## What did you implement:

Closes: #161

I compared the results of creating a container via the command line and via shiplift (confirming that the ports column is populated correctly). I also curl'd against the REST API to retrieve the json from the running containers to compare before and after.

Added `ExposedPorts` mapping when creating a container.

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -582,9 +582,9 @@ impl ContainerOptionsBuilder {
         exposedport.insert("HostPort".to_string(), hostport.to_string());
 
         /* The idea here is to go thought the 'old' port binds
-         * and to apply them to the local 'binding' variable,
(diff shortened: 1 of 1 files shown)


softprops:
@gwihlidal this looks good. Could you run `cargo +nightly fmt --all` and push up the result? I use [rustfmt](https://github.com/softprops/shiplift/blob/master/CONTRIBUTING.md#configuring-rustfmt) to keep a consistent formatting of source code.

gwihlidal:
I applied cargo fmt as requested. Is it possible to merge this and then push a new version of shiplift to cargo?

Thanks!!

softprops:
This is long over due. I'll try go commit to that tomorrow

softprops:

Issue #163, opened 2019-04-25
Support for bind mounts?


keirlawson:
So far as I can see Shiplift does not support bind mounts (https://docs.docker.com/storage/bind-mounts/) only named volumes, is this the case? Would it be possible to add bind mounts support if it is missing?

schrieveslaach:
I think that you can just use the path to file/directory you want to mount:

```
docker
    .containers()
    .create(&ContainerOptions::builder(image.as_ref()).volumes(vec!["/path/to/file:/path/in/container"]).build())
```

keirlawson:
Ah yep confirmed you can, had assumed that option was for named volumes, will open a PR to add some clarificatory documentation.

PR #164, opened 2019-04-26
Add explanative sentence for volumes option

keirlawson:
## What did you implement:

Added a documentation for the `volumes` option.

Closes: #163 

Generated documentation locally

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -564,6 +564,7 @@ impl ContainerOptionsBuilder {
         self
     }
 
+    /// Specify any bind mounts, taking the form of `/some/host/path:/some/container/path`
     pub fn volumes(
         &mut self,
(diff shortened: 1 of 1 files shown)


softprops:
Thanks @keirlawson

src/builder.rs
```diff
@@ -564,6 +564,7 @@ impl ContainerOptionsBuilder {
         self
     }
 
+    /// Specify any bind mounts, taking the form of `/some/host/path:/some/container/path`
```
schrieveslaach: I think `volumes` is not restricted to bind mounts. You can use also use named volumes. For example, you could use `volumes(vec!["myvol2:/app"])` (see [here](https://docs.docker.com/storage/volumes/))

PR #170, opened 2019-05-18
Switch to vendored openssl

keirlawson:
Switch to vendored openssl to ease building on windows.


softprops:
@keirlawson  can you walk me though what this does so I know what to put in the changelog?

keirlawson:
This should mean that we build against OpenSSL bundled via a crate rather than the system OpenSSL lib, which can be painful to set up on windows. See https://docs.rs/openssl/0.10.23/openssl/#vendored

PR #171, opened 2019-05-22
copy_file_into creates sub-directories if required

schrieveslaach:
## What did you implement:

The `copy_file_into` function required that the parent directory of the file to copy exists in the container. This PR ensures that the required path will be created automatically. 

Tested with the example `examples/containercopyinto.rs` and tested with [PREvant](https://github.com/aixigo/PREvant) by [this commit](https://github.com/aixigo/PREvant/pull/7/commits/4c7e25685219a0f3521be70267d893ab2e1c23f9).

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -539,8 +539,7 @@ impl<'a, 'b> Container<'a, 'b> {
     /// Copy a byte slice as file into (see `bytes`) the container.
     ///
     /// The file will be copied at the given location (see `path`) and will be owned by root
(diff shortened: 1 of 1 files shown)


softprops:
Thanks for the patch. Would you mind running rust fmt once and push up the result https://github.com/softprops/shiplift/blob/master/CONTRIBUTING.md#configuring-rustfmt

schrieveslaach:
Sure, sorry about that. :v:

PR #181, opened 2019-07-05
fixed small error

0xflotus:
I only fixed a typo


softprops:
Travis ci problem is related to a change in rustfmt. I'll fix that separately. Thanks for the patch @0xflotus!

Issue #186, opened 2019-08-07
Add Chrono Feature


schrieveslaach:
The crate [`chrono`](https://docs.rs/chrono/0.4.7/chrono/) supports parsing of date times in RFC3339 format and the Docker API uses this format e.g. to desribe the start time of a container. I would be vary handy if `shiplift` would have an optional feature to deserialize these fields as `DateTime<FixedOffset>`. For example, this PR aixigo/PREvant#13 would benefit from direct support of chrono.

softprops:
I would be open to a pr to add this feature

schrieveslaach:
I can make the PR. Should one question up front. Do you think it should be a feature toggle or do you think chrono should be supported by default?

softprops:
I would say a feature toggles that's on by default with the option to turn off would be generally ideal but I'd also be find without a toggle since chrono is pretty ubiquitous in the ecosystem

schrieveslaach:
@softprops, there is a PR implementing deserialization with `chrono`.

PR #190, opened 2019-09-03
feat: use chrono to deserialize date times

schrieveslaach:
## What did you implement:

Add chrono as a feature to deserialize date times 

Closes: #186 

Tested with all examples and uni tests


softprops:
Thanks @schrieveslaach!

Issue #192, opened 2019-09-13
Detect tls stream and use `https` scheme instead of provided `tcp`


LucioFranco:
When attempting to use shiplift in a circleci environment with hyper, it must use `tcp` to connect to the remote docker daemon. This requires tls via the docker docs. The issue arises here https://github.com/softprops/shiplift/blob/master/src/transport.rs#L166 when the `tcp://` scheme is copied. The real problem is that hyper's version checking requires the use of `https` instead of `tcp` directly. Thus the fix here is to replace the scheme only when encrypted.

Reference failing tests in https://github.com/timberio/vector/pull/787

elihunter173:
Should this issue still be open?

LucioFranco:
I think its been fixed!

PR #193, opened 2019-09-13
Replace host str to use https if encrypted

LucioFranco:
During our testing in https://github.com/timberio/vector/pull/787 we found that there is an issue with hyper connecting to a tls connected docker engine that provided a host url of `tcp://...`. I found the way to fix this was to replace the `tcp` scheme with `https`. This seems to work for when encrypted `tcp` streams are active. The error hyper produces was a version parse error which seemed odd. This seemed to do the trick.

Related to #192

### DIFF
--- a/examples/volumes.rs
+++ b/examples/volumes.rs
@@ -12,7 +12,7 @@ fn main() {
                 println!("volume -> {:#?}", v)
             }
         })
-        .map_err(|e| eprintln!("Error: {}", e));;
+        .map_err(|e| eprintln!("Error: {}", e));
 
     tokio::run(fut);
 }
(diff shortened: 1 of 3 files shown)


softprops:
Could we add a little code comment above these  lines to capture the context. I could see my future self coming back to this a year from now and asking "why?"

LucioFranco:
@softprops Sorry for the delay I've added a comment, if we could get a release that would be very appreciated, let me know what I can do to help!

softprops:
Thanks. I'm gonna shoot for at least this weekend.

LucioFranco:
Thank you Doug <3

PR #194, opened 2019-09-15
Use 'dyn' since trait objects without an explicit 'dyn' are deprecated

Atul9:
## What did you implement:
->  I found that trait objects without an explicit dyn are deprecated - https://travis-ci.org/softprops/shiplift/jobs/580271710, and I followed the suggestion given by the compiler to fix this.

Closes: #xxx

## What (if anything) would need to be called out in the CHANGELOG for the next release: Since this is just a compiler warning fix should we add it to CHANGELOG?

### DIFF
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -79,7 +79,7 @@ impl StdError for Error {
         }
     }
 
-    fn cause(&self) -> Option<&StdError> {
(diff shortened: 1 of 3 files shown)


softprops:
Thanks for grabbing this @Atul9

PR #196, opened 2019-09-29
Correctly rename `endpoint_id` field

pitkley:
## What did you implement:

The `endpoint_id` field is stylized as `EndpointID` in the Docker API when returning container-details for a network ([1]).

[1]: https://docs.docker.com/engine/api/v1.24/#35-networks

When previously I called `.inspect` on a Docker network, it failed with the (Serde) error:

```
missing field `EndpointId`
```

After changing the rename-definition of the `endpoint_id` field in `NetworkContainerDetails` it now works correctly.


softprops:
sry for the delay @pitkley. thanks for the patch

PR #197, opened 2019-10-02
Don't vendor OpenSSL except on Windows

bossmc:
## What did you implement:

This relates to #170 where shiplift switched to using vendored openssl to assist windows builds.  Unfortunately I believe that change went too far the other way, using the system-wide OpenSSL is much more common on *nix systems, and the vendoring gets in the way of other tooling (e.g. building an RPM from such a binary leads to problems, as the RPM will try to "depend" on the vendored shared object, which will not be resolvable from the repo).

This change enables the `vendored` feature only on windows builds, and reverts to the default behaviour (searching for a system-wide version using pkg-config if present) on other systems (mac, linux...)

I considered an alternative option, that of exposing a `vendored-ssl` feature that enables the `openssl/vendored` feature, which can be enabled if desired by builds, but that makes windows builds harder, which feels like a step backwards.

(cross) Compiled for linux and windows.

Clarifying the notes from #170 to be clear that they only apply to windows builds.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,6 +30,7 @@ hyper-openssl = { version = "0.7", optional = true }
 hyperlocal = { version = "0.6", optional = true }
 log = "0.4.6"
 mime = "0.3.13"
+openssl = { version = "0.10", optional = true }
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
 tar = "0.4"
@@ -38,11 +39,6 @@ tokio-codec = "0.1"
 tokio-io = "0.1"
 url = "1.7"
 
-[target.'cfg(windows)'.dependencies]
-openssl = { version = "0.10", optional = true, features = ["vendored"] }
-[target.'cfg(not(windows))'.dependencies]
-openssl = { version = "0.10", optional = true }
-
 [dev-dependencies]
 env_logger = "0.6"
 
@@ -50,3 +46,4 @@ env_logger = "0.6"
 default = ["chrono", "unix-socket", "tls"]
 unix-socket = ["hyperlocal"]
 tls = ["openssl", "hyper-openssl"]
+vendored-ssl = ["tls", "openssl/vendored"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,7 +30,6 @@ hyper-openssl = { version = "0.7", optional = true }
 hyperlocal = { version = "0.6", optional = true }
 log = "0.4.6"
 mime = "0.3.13"
-openssl = { version = "0.10", optional = true, features = ["vendored"] }
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
 tar = "0.4"
@@ -38,6 +37,11 @@ tokio = "0.1"
 tokio-codec = "0.1"
 tokio-io = "0.1"
 url = "1.7"
+
+[target.'cfg(windows)'.dependencies]
+openssl = { version = "0.10", optional = true, features = ["vendored"] }
+[target.'cfg(not(windows))'.dependencies]
+openssl = { version = "0.10", optional = true }
 
 [dev-dependencies]
 env_logger = "0.6"
--- a/src/transport.rs
+++ b/src/transport.rs
@@ -221,7 +221,8 @@ impl Transport {
             Transport::Tcp { .. } => (),
             #[cfg(feature = "tls")]
             Transport::EncryptedTcp { .. } => (),
-            _ => panic!("connection streaming is only supported over TCP"),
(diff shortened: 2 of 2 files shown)


Cargo.toml
```diff
@@ -39,6 +38,11 @@ tokio-codec = "0.1"
 tokio-io = "0.1"
 url = "1.7"
 
+[target.'cfg(windows)'.dependencies]
```
richardwhiuk: This should be `cfg(target_os = "windows")` 

See https://github.com/sfackler/rust-native-tls/blob/master/Cargo.toml#L20 for an example that works.

richardwhiuk:
It seems like this should work, but it actually doesn't - see https://github.com/rust-lang/cargo/issues/1197#issuecomment-231584705

😞

bossmc:
Sounds like adding a feature is a better fix then.

softprops:
Just catching up on the state of this. I don't have a great way to test on windows at the moment.  Are there any pending changes?

bossmc:
@softprops - There's still some more work to do, the change I've made so far is being ignored by Cargo, so I'm putting together a smaller version of the change.  Should be with you today.

bossmc:
Ok, updated PR, new strategy - expose a `vendored-ssl` feature on shiplift that activates the `openssl/vendored` feature on openssl.

Windows:
```
c:\src\shiplift>cargo t --features vendored-ssl --no-default-features
...
test result: ok. 10 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
```

Linux:
```
$ cargo t
...
test result: ok. 10 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
$ ldd target/debug/deps/shiplift-5d6a9fe8f830c081
...
	libssl.so.1.1 => /usr/lib/x86_64-linux-gnu/libssl.so.1.1 (0x00007fccd9fa1000)
...
$ cargo t --features vendored-ssl
...
test result: ok. 10 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
$ ldd target/debug/deps/shiplift-792b6a711bfb1b6c
	linux-vdso.so.1 (0x00007f7568604000)
	libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007f7567e7e000)
	librt.so.1 => /lib/x86_64-linux-gnu/librt.so.1 (0x00007f7567c76000)
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f7567a57000)
	libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007f756783f000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f756744e000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f75683de000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f75670b0000)
```

src/transport.rs
```diff
@@ -221,7 +221,8 @@ impl Transport {
             Transport::Tcp { .. } => (),
             #[cfg(feature = "tls")]
             Transport::EncryptedTcp { .. } => (),
-            _ => panic!("connection streaming is only supported over TCP"),
+            #[cfg(feature = "unix-socket")]
+            Transport::Unix { .. } => panic!("connection streaming is only supported over TCP"),
```
bossmc: This change prevents an "unreachable branch" warning on builds without unix-socket support (e.g. on windows)

softprops:
thanks for your work on this @bossmc I'm going to try and cut a release this weekend

thomaseizinger:
For what it is worth, at least the currently released version doesn't work on Windows because `hyper-openssl` [only actually establishes an SSL connection if the scheme of the URI is `https`](https://github.com/sfackler/hyper-openssl/blob/master/src/lib.rs#L173).

However, the `DOCKER_HOST` variable on Windows (at least with Docker Toolbox), is something like: `tcp://192.168.1.100:2367`. We had to [patch around this behaviour](https://github.com/comit-network/create-comit-app/blob/51e6e8f444bae080abedca24795bb0a6c4144a6f/src/docker/mod.rs#L189-L213) to actually make it work.

The new Docker for Windows (using HyperV) is another whole different story because it doesn't set the `DOCKER_HOST` variable. I haven't yet figured out how to make that work without disabling TLS.

thomaseizinger:
Don't mind me, I just saw that there was another PR merged today about this: https://github.com/softprops/shiplift/pull/193/files

bossmc:
Cool, thanks @softprops!

bossmc:
I've updated the notes for the CHANGELOG in the main PR description.

PR #198, opened 2019-10-13
Add 'publish' method to expose a port

Miaxos:
# What did you implement:
```
ContainerOptionsBuilder::publish();
```

I added the publish function to only expose port on the container instead of mapping ports.

# Context:
I got an automated nginx_proxy (https://github.com/jwilder/nginx-proxy) that'll detect automatically if a container is added to his network and for that, you only need to expose ports of the container with a `docker run --expose 5000 something` and if multiples, specify it with env variables

But with shiplift, I couldn't only expose a port, I had to do the mapping.

# How did you verify your change: 
By testing and with units tests.

# What (if anything) would need to be called out in the CHANGELOG for the next release:
publish method added to publish container's ports.

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -685,6 +685,40 @@ impl ContainerOptionsBuilder {
         self
     }
 
+    /// Publish a port in the container without assigning a port on the host
+    pub fn publish(
+        &mut self,
+        srcport: u32,
+        protocol: &str,
+    ) -> &mut Self {
+        /* The idea here is to go thought the 'old' port binds
+         * and to apply them to the local 'exposedport_bindings' variable,
+         * add the bind we want and replace the 'old' value */
+        let mut exposed_port_bindings: HashMap<String, Value> = HashMap::new();
+        for (key, val) in self
+            .params
+            .get("ExposedPorts")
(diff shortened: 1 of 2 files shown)


src/builder.rs
```diff
@@ -1721,7 +1775,9 @@ mod tests {
             .server_address("https://example.org")
             .build();
         assert_eq!(
-            base64::encode(r#"{"username":"user_abc","password":"password_abc","email":"email_abc","serveraddress":"https://example.org"}"#),
+            base64::encode(
```
Miaxos: rustfmt doing

src/builder.rs
```diff
@@ -1676,7 +1730,7 @@ mod tests {
         assert_eq!(
             r#"{"HostConfig":{"RestartPolicy":{"MaximumRetryCount":10,"Name":"on-failure"}},"Image":"test_image"}"#,
             options.serialize().unwrap()
-       );
+        );
```
Miaxos: rustfmt doing

Miaxos:
I'll write the issue, then add the PR

softprops:
thanks @Miaxos . looks great

PR #204, opened 2019-12-06
port build to gh actions

softprops:
## What did you implement:

Ports build to GitHub actions. It has anecdotally provides a bit more in the way of putting my projects on auto pilot.

Closes: #xxx


src/builder.rs
```diff
@@ -676,7 +676,7 @@ impl ContainerOptionsBuilder {
         // Replicate the port bindings over to the exposed ports config
         let mut exposed_ports: HashMap<String, Value> = HashMap::new();
         let empty_config: HashMap<String, Value> = HashMap::new();
-        for (key, _) in &port_bindings {
+        for key in port_bindings.keys() {
```
softprops: passes clippy checks

Issue #183, opened 2019-07-15
http::Uri instead of url::Url


mfirry:
- Crate version: 0.5
- OS: macOS Mojave (10.14.5)
- Output of running `docker version` on the command line:
```
Client: Docker Engine - Community
 Version:           18.09.2
 API version:       1.39
 Go version:        go1.10.8
 Git commit:        6247962
 Built:             Sun Feb 10 04:12:39 2019
 OS/Arch:           darwin/amd64
 Experimental:      false

Server: Docker Engine - Community
 Engine:
  Version:          18.09.2
  API version:      1.39 (minimum version 1.12)
  Go version:       go1.10.6
  Git commit:       6247962
  Built:            Sun Feb 10 04:13:06 2019
  OS/Arch:          linux/amd64
  Experimental:     false
```

Instead of
```
use url::Url;

let docker = Docker::host(Url::parse("http://yourhost").unwrap());
```

I think one should do
```
use http::Uri;

let docker = Docker::host("http://yourhost".parse::<Uri>().unwrap());
```

schrieveslaach:
I think that the [url](https://docs.rs/url/1.7.2/url/) is a much smaller library and should be a perfect match here because it is only required to parse an URL. The [http](https://docs.rs/http/0.1.17/http/) provides much more type types that are not required for the use cases of shiplift.

mfirry:
I'm confused now. I opened the issue because for me `Url` wasn't working and I had errors saying I should provide a `Uri`. Am I missing something?

schrieveslaach:
Ah, okay, I messed up the issue description and I thought you suggested to use `http::Uri` instead `url::Url`. I agree that `url::Url` is the much better choice.

danieleades:
@schrieveslaach `http` is a transitive dependency of `hyper` anyway, no?

schrieveslaach:
Yes, [`hyper` relies on `http`](https://github.com/hyperium/hyper/blob/master/Cargo.toml#L27)

danieleades:
What I mean is, it doesn't matter whether it has extra types or functionality you don't need, since it's statically linked already, so no matter how lightweight `Url` is, it's duplicated functionality. Is it more ergonomic for parsing url strings in some way than the `http` types?

schrieveslaach:
[...]

PR #207, opened 2019-12-10
Replace invalid example in README with cargo example

thomaseizinger:
## What did you implement:

The example in the README is outdated and no longer compiles. To prevent this from happening again, this patch removes the example entirely and instead introduces a new example `custom_host` that will be checked during `cargo test`.

Closes: #183.

-

-

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,26 +15,6 @@ shiplift = "0.6"
 
 ## usage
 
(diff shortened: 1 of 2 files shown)


softprops:
thanks @thomaseizinger !

PR #211, opened 2019-12-17
Make image and container option builder interfaces consistent

colvin:
## What did you implement:

`builder::ContainerListOptionsBuilder::all()` takes no arguments, assuming a true
value. However, `builder::ImageListOptionsBuilder::all()` accepts a boolean.

To make these consistent, the latter no longer accepts a boolean and,
like the former, assumes true.

`cargo test`

`shiplift::builder::ImageListOptionsBuilder::all()` no longer accepts an argument and always sets the option to true.

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -1298,11 +1298,8 @@ impl ImageListOptionsBuilder {
         self
     }
 
-    pub fn all(
-        &mut self,
-        a: bool,
-    ) -> &mut Self {
-        self.params.insert("all", a.to_string());
+    pub fn all(&mut self) -> &mut Self {
+        self.params.insert("all", "true".to_owned());
         self
     }
 


src/builder.rs
```diff
@@ -1298,11 +1298,8 @@ impl ImageListOptionsBuilder {
         self
     }
 
-    pub fn all(
-        &mut self,
-        a: bool,
-    ) -> &mut Self {
-        self.params.insert("all", a.to_string());
+    pub fn all(&mut self) -> &mut Self {
```
softprops: What would you think about flipping this around and adding a bool parameter to the other option builder? Since it's an option in the api to provide an explicit false I don't see an advantage to hiding that from users.

colvin: Correct me if I'm wrong, but I would think that, generally, the builder API design concept implies that calling a method like `all()` would positively apply the attribute to which it corresponds.

The Docker API's convention is that the `all` option is off by default ([/containers/json](https://docs.docker.com/engine/api/v1.24/#31-containers), [/images/json](https://docs.docker.com/engine/api/v1.24/#32-images)), but can be enabled upon request.

With those two in mind, I would think that assuming `true` for the `all()` methods of the builder APIs would be the logical behavior.

softprops: The default it's true but the api docker exposes takes an argument. I was throwing it out there just to see what you thought.

PR #213, opened 2019-12-17
Expand rep::Container fields

colvin:
## What did you implement:

Add missing fields to `rep::Container`:

- `image_id` (`ImageID`, API version 1.21)
- `state` (`State`, API version 1.23)

`cargo test`


softprops:
Thanks @colvin

Issue #214, opened 2019-12-18
Request to add the 'PublishAllPorts' option to the ContainerOptionsBuilder


tylerhjones:
I am using dockertest-rs in my integration tests. After trying to manage locks and host:port mappings, the maintainer pointed out to me that docker now has a 'PublishAllPorts' option available. This option will assign an available host port at random for each exposed port on the container. The ContainerOptionsBuilder does not currently support this option.

[Documented Option](https://docs.docker.com/engine/api/v1.39/#operation/ContainerCreate), search for 'PublishAllPorts'

I would like to add this to the ContainerOptionsBuilder by introducing code such as

```rust
    pub fn publish_all_ports(
        &mut self,
    ) -> &mut Self {
        self.params.insert("HostConfig.PublishAllPorts", json!(true));
        self
    }
```

```rust
    /// Test container option PublishAllPorts
    #[test]
    fn container_options_publish_all_ports() {
        let options = ContainerOptionsBuilder::new("test_image")
            .publish_all_ports()
            .build();

        assert_eq!(
            r#"{"HostConfig":{"PublishAllPorts":true},"Image":"test_image"}"#,
            options.serialize().unwrap()
        );
    }
```

softprops:
Looks great @tylerhjones

PR #215, opened 2019-12-18
Update the ContainerOptionsBuilder with new publish all ports option

tylerhjones:
The following change is to provide the configuration option
`PublishAllPorts` which will map all exposed ports on a container to
random and available ports on the host machine.

This configuration option is documented
[here](https://docs.docker.com/engine/api/v1.39/#operation/ContainerCreate).

CHANGELOG.md is not updated, because I do not think there are any breaking changes.

## What did you implement:

Closes: https://github.com/softprops/shiplift/issues/214

I have added a test to the builder; builder::tests::container_options_publish_all_ports
Ran cargo test with the following results.
```
running 12 tests
test builder::tests::container_options_publish_all_ports ... ok
test builder::tests::container_options_nested ... ok
test builder::tests::container_options_host_config ... ok
test builder::tests::container_options_env ... ok
test builder::tests::container_options_simple ... ok
test builder::tests::container_options_restart_policy ... ok
test builder::tests::container_options_publish ... ok
test builder::tests::registry_auth_password_all ... ok
test builder::tests::registry_auth_password_simple ... ok
test builder::tests::logs_options ... ok
test builder::tests::registry_auth_token ... ok
test builder::tests::container_options_expose ... ok

test result: ok. 12 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out

   Doc-tests shiplift

running 1 test
test src/lib.rs -  (line 5) ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
```

 Nothing I think?

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -642,6 +642,14 @@ impl ContainerOptionsBuilder {
         self
     }
 
+    /// enable all exposed ports on the container to be mapped to random, available, ports on the host
+    pub fn publish_all_ports(
+        &mut self,
+    ) -> &mut Self {
+        self.params.insert("HostConfig.PublishAllPorts", json!(true));
+        self
+    }
+
     pub fn expose(
         &mut self,
         srcport: u32,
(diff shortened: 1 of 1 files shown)


tylerhjones:
I have a few concerns I was not sure how to go about testing.
1) Docker version support
  This feature is documented in the earliest version documented here https://docs.docker.com/engine/api/version-history/ (1.18), and is in the latest with no mentioned change in the change logs, but I am not sure about possibly earlier versions not listed on that page.

2)
I did not test possible conflicts if this `PublishAllPorts` and the port mapping are provided.
I assumed this to be the responsibility of docker, but maybe this tool wants to do some checks before hand in the builder for possibly conflicting configurations? (again assumes it would even be a conflict)

softprops:
Thanks for the addition @tylerhjones

Issue #216, opened 2019-12-29
Implement Deserialize for ContainerOptions


zicklag:
Is it possible to have [`ContainerOptions`](https://docs.rs/shiplift/0.6.0/shiplift/builder/struct.ContainerOptions.html) implement `Deserialize`? In my project I am intending to be able to serialize and deserialize a struct that contains a field of type `ContainerOptions`.

zicklag:
After looking into the implementation it seems like the easiest route would be to derive Serialize and Deserialize on the `ContainerOptionsBuilder`. Then I could store the `ContainerOptionsBuilder` instead of the `ContainerOptions`. Does that sound like a good solution to you? Would you be open to a PR?

zicklag:
OK, I opened a draft PR: #217. Please tell me what you think of it. :smiley:

PR #217, opened 2019-12-30
Make ContainerOptions Deserializable

zicklag:
## What did you implement:

This implements Deserialize on `ContainerOptions`. I switched the keys in the `ContainerOptions.params` hash map to owned strings instead of `&'static str`. Technically these could be borrowed from the deserialized value, but I think that would be cumbersome to use.

Closes: #216

Not at all yet ( this is work-in-progress ). I'll be testing it soon.

Just a note that Deserialize is now implemented for `ContainerOptions`.

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -1,7 +1,7 @@
 //! Interfaces for building various structures
 
 use crate::{errors::Error, Result};
-use serde::Serialize;
+use serde::{Serialize, Deserialize};
 use serde_json::{self, json, map::Map, Value};
 use std::{
     cmp::Eq,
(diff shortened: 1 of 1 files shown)


zicklag:
OK this is turning into a little bit more than making `ContainerOptions` serializable. After trying to use the changes in my project I realized that it made more sense for me to make `ContainerOptionsBuilder` deserializable because I need to be able to create the builder, write the builder to disk, read the builder from disk, and modify the builder whenever I want. There is no good way to modify `ContainerOptions` so I had to use the builder instead.

Additionally I need to be able to change the container image after the builder has been instantiated, so I created a function for that.

I am also going to need the ability to remove exposed ports, so I will add that to this PR as well if you are OK with that.

softprops (approved):
Thanks for the change

softprops:
Sorry about the delay on this.

 For a long while all of my gh motifs were going to my email spam folder so I likely missed this question before. I think this is a good change that adds some flexibility that doesn’t exist today. This is one of my earlier crates so I was still figuring out what good looked like. If I had more time I’d be doing more refactoring.

I really appreciate pulls like yours

zicklag:
No problem, glad it's useful. :) I think I was going to need this for a project and then didn't end up needing it.

I've personally written code, and then gone back to it and been like, "oh, I wish I would have known what I know now when I wrote this, it would have been a lot nicer". :slightly_smiling_face:

softprops:
Ok. Let me know if you want to move forward with this. If not let's revisit this when inspiration strikes again :)

PR #220, opened 2020-02-18
Specify user when creating container

tofay:
## What did you implement:
A User can now be specified when creating a container, per https://docs.docker.com/engine/api/v1.30/#operation/ContainerCreate

Ran tests, and tested locally.

Nothing.


bossmc:
@softprops - is there any chance of a release with this change in it?  Or is there something else you're trying to get in to the next cut?

PR #223, opened 2020-03-18
Update errors to match current std::error::Error trait

colvin:
## What did you implement:

Update the internal `Error` type to the latest `std::error::Error` trait:
- `description` is deprecated, and with the 1.42 stable release it is now generating warnings
- `cause` is deprecated and replaced with `source`

`cargo test`


softprops:
thanks @colvin

PR #232, opened 2020-09-01
Implement multiple chunk streams in build function for images

lidin:
## What did you implement:

Streaming data in build function for images was no longer supporting multiple chunks, and thus intermittently fails with a Serde JSON parsing error, as described in #143. This PR fixes that for the async/await syntax support.

Run the imageebuild example specifying the path of a Dockerfile:

```shell
cargo run --example imagebuild <path>
```

Running this before this PR reveals the Serde JSON error, and running it after this PR shows that the errors are no longer emitting.

- Re-add support for multiple chunks

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -170,9 +170,18 @@ impl<'a> Images<'a> {
                     None::<iter::Empty<_>>,
                 );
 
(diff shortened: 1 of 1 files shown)


softprops:
nice work @idin

PR #233, opened 2020-09-05
Run examples

jcaromiq:
## What did you implement:
Add in cargo file examples section to run all examples

run examples with cargo --example

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -49,3 +49,135 @@ default = ["chrono", "unix-socket", "tls"]
 unix-socket = ["hyperlocal"]
 tls = ["openssl", "hyper-openssl"]
 vendored-ssl = ["tls", "openssl/vendored"]
+
+[[example]]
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -49,3 +49,135 @@ default = ["chrono", "unix-socket", "tls"]
 unix-socket = ["hyperlocal"]
 tls = ["openssl", "hyper-openssl"]
 vendored-ssl = ["tls", "openssl/vendored"]
+
+[[example]]
```
softprops, line 53: Thanks for the contribution.

I don't believe these are actually needed unless we were to name one of the examples something other than the name of the file.

softprops: Cargo uses the convention of resolving an example by file name by default

jcaromiq: I didn't know! 
I close the PR
Thanks

PR #237, opened 2020-10-16
ContainerOptionsBuilder::env() should not get owned Vec

matthiasbeyer:
This type signature makes it impossible to dynamically construct a
`Vec<String>` of environment variables and pass it to the builder.

This patch tries to fix that by making the function only take a
reference to a slice containing the environments.

Added a test.

`ContainerOptionsBuilder::env()` signature changed

---

Please take this as a proposal. It is most certainly not the nicest solution for the problem described above.

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -854,10 +854,13 @@ impl ContainerOptionsBuilder {
         self
     }
 
-    pub fn env(
+    pub fn env<E, S>(
         &mut self,
-        envs: Vec<&str>,
-    ) -> &mut Self {
+        envs: E,
+    ) -> &mut Self
+        where S: AsRef<str> + Serialize,
+              E: AsRef<[S]> + Serialize
+    {
         self.params.insert("Env", json!(envs));
         self
     }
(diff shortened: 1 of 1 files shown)


matthiasbeyer:
A possible (user-side) solution would be

```
let f: Vec<&str> = v.iter().map(|s| s.deref()).collect();
```

But this is hella ugly IMO.

elihunter173:
Is there any reason to not just use slices like so?

```rust
pub fn env(&mut self, envs: &[&str]) -> &mut Self { ... }
```

matthiasbeyer:
This is what this patch does, but with more flexibility even.

src/builder.rs
```diff
@@ -854,10 +854,13 @@ impl ContainerOptionsBuilder {
...
     }
 
-    pub fn env(
+    pub fn env<E, S>(
         &mut self,
-        envs: Vec<&str>,
-    ) -> &mut Self {
+        envs: E,
+    ) -> &mut Self
+        where S: AsRef<str> + Serialize,
```
softprops, line 861: Nice change. Both ergonomically and for flexibility

softprops (approved):

Issue #166, opened 2019-05-03
Can't attach to running containers


keirlawson:
```rust
use shiplift::{Docker};
use tokio::runtime::current_thread::Runtime;
use tokio::prelude::Future;
use futures::stream::Stream;

fn main() {
    let docker = Docker::new();

    let mut rt = Runtime::new().unwrap();

    let cont = docker.containers().get("2642f3eabd5f");

    let fut = cont.attach()
        .and_then(|multiplexed|
            multiplexed.for_each(|c| {
                println!("{:?}", c);
                Ok(())
            })
        );

    rt.block_on(fut).unwrap();
}
```

This program gives the following error when run:

```
thread 'main' panicked at 'connection streaming is only supported over TCP', /Users/keirlawson/.cargo/registry/src/github.com-1ecc6299db9ec823/shiplift-0.5.0/src/transport.rs:224:18
note: Run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
```

But it seems to me that it should be possible to attach across the default unix socket?

- Crate version: 0.5.0
- OS: Mac OS X
- Output of running `docker version` on the command line:
```
docker version                                                                                                                                                                                                 
Client: Docker Engine - Community
 Version:           18.09.2
 API version:       1.39
 Go version:        go1.10.8
 Git commit:        6247962
 Built:             Sun Feb 10 04:12:39 2019
 OS/Arch:           darwin/amd64
 Experimental:      false

Server: Docker Engine - Community
 Engine:
  Version:          18.09.2
  API version:      1.39 (minimum version 1.12)
  Go version:       go1.10.6
  Git commit:       6247962
  Built:            Sun Feb 10 04:13:06 2019
  OS/Arch:          linux/amd64
  Experimental:     false
```

PR #238, opened 2020-11-10
Enable streaming results over UNIX sockets

fooker:
## What did you implement:

This simply removes the the restriction for streaming on UNIX domain sockets. It works for my use cases and I haven't found any documentation that the mechanism differs from streaming over TCP.

Closes: #166

Using `container.attach()` over a local docker instance without TCP enabled.

*nothing*

### DIFF
--- a/src/transport.rs
+++ b/src/transport.rs
@@ -236,14 +236,6 @@ impl Transport {
     where
         B: Into<Body>,
     {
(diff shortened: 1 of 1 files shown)


softprops:
Nice catch. The blocker must have been removed at some point

PR #239, opened 2020-11-11
Implement upload of tar to container

fooker:
## What did you implement:

Added a `Container::copy_to` method which allows uploading a pre-existing tarball to the container.

Calling the new method with an archive generated by the `tar` crate.

*nothing*

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -636,7 +636,7 @@ impl<'a> Container<'a> {
 
     /// Copy a tarball (see `body`) to the container.
     ///
-    /// Tha tarball will be copied to the container and extracted at the given location (see `path`).
+    /// The tarball will be copied to the container and extracted at the given location (see `path`).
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -630,16 +630,28 @@ impl<'a> Container<'a> {
...
         let data = ar.into_inner().unwrap();
 
-        let body = Some((data, "application/x-tar".parse::<Mime>().unwrap()));
+        self.copy_to(Path::new("/"), data.into()).await?;
+        Ok(())
+    }
 
+    /// Copy a tarball (see `body`) to the container.
+    ///
+    /// Tha tarball will be copied to the container and extracted at the given location (see `path`).
```
elihunter173, line 639:
```suggestion
    /// The tarball will be copied to the container and extracted at the given location (see `path`).
```

PR #240, opened 2020-12-02
Allow `Image::pull` to handle chunks with multiple JSON values

tofay:
## What did you implement:

Move logic from `Image::build` for converting a stream of chunks into a stream of values to a new `stream_post_into_values` function, and use that in `Image::pull`.

This fixes a bug on WSL2 where shiplift errors when calling `Image::pull` on an image that has already been pulled. In this case the response from the docker daemon contains two JSON values in a single chunk.

Successfully ran the image pull example twice in succession with the same argument.
Nothing.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1131,6 +1131,9 @@ impl Docker {
         Ok(serde_json::from_str::<T>(&string)?)
     }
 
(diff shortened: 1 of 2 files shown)


softprops (approved):
Looks good. Sorry for the delay

PR #242, opened 2020-12-04
Move error implementation to use thiserror

matthiasbeyer:

`cargo check --all --all-features`

* errors are implemented using [thiserror](https://crates.io/crates/thiserror)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # 0.7.0
 
 * async-await support [#229](https://github.com/softprops/shiplift/pull/229)
+* errors are now implemented using `thiserror` [#242](https://github.com/softprops/shiplift/pull/242)
 
 # 0.6.0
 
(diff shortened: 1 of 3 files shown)


matthiasbeyer:
I will squash if you approve!

softprops:
Thanks for your work on this.

however, I tend to err on the side of longer term reductions in maintenance. 

I’ve been bitten more than one by claiming short term maintenance wins by using an error derivation crate but then as the ecosystem around those change I’ll find myself having to take on hidden maintenance up updating to the new latest and greatest error derivation crate. 

The cost maintenance of impl errors once and never having to think about them or keeping up with a dependency or error ecosystem trends is actually quite liberating

This change might seem like it would solve a maintenance problem but in practice error types tend to change the least compared to other code so it’s a optimization for a problem I don’t have today as well as introduces a new maintenance on a dependency I didn’t have before.

matthiasbeyer:
I agree. :smile:

PR #245, opened 2021-01-12
Fix X-Registry-Auth header

valff:
Current implementation of `RegistryAuth` doesn't work, because it should be encoded in url_safe base64 variant.

Relevant docs: https://docs.docker.com/engine/api/v1.41/#section/Authentication


softprops (approved):
I've catch!

softprops:
*nice

PR #246, opened 2021-01-13
Upgrade all dependencies

micromaomao:
This pull request upgrades all dependencies to their latest version and fix incompatible code. Specifically, this updates tokio from 0.2 to 1.0, which makes shiplift usable in newer projects.

## What did you implement:

* Changed version strings in Cargo.toml.
* Implemented a special case for `unix:///` URLs, since `"unix:///var/run/docker.sock".parse::<hyper::Uri>()` will now return an error saying the authority can't be empty.
* Created unit test for this special case.
* For some reason there is an empty `/lib.rs`. This pull request also removed it.

* `cargo test --no-default-features` pass
* `cargo test` pass
* Tried `cargo run --example info` with `DOCKER_HOST` set to `unix:///...`, `http://...` and `tcp://...` URLs.
* Verified that creating containers, starting and attaching to them still works, since I'm using this on my own project.

No changes to the interface is made, however this would probably require a minor version bump, since updating tokio will force dependent of this project to also update their tokio as well.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,31 +18,31 @@ coveralls = { repository = "softprops/shipflit" }
 maintenance = { status = "actively-developed" }
 
 [dependencies]
-base64 = "0.11"
+base64 = "0.13"
 byteorder = "1.3"
-bytes = "0.4"
+bytes = "1.0"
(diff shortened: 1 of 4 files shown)


softprops:
Excellent!

PR #250, opened 2021-01-26
Make tarball a generic argument.

### LINKED ISSUE #249: Import Stream is not Send.
It would be nice to either use a generic `pub fn import<R: Read>(self, mut tarball: R)` or make the bound `Box<dyn Read + Send>`.

```rust
pub fn import(
        self,
        mut tarball: Box<dyn Read>,
) -> impl Stream<Item = Value, Error = Error> {
```

dvc94ch:
Closes #249


dvc94ch:
Don't need this anymore as we aren't using shiplift

Issue #219, opened 2020-02-05
Get Exec Exit Code


zicklag:
I'm using the `exec` function to execute a command in a container, and I need to be able to find out what the exit code of the command was. It looks like there isn't any way to do so with the current API.

softprops:
Hi @zicklag I would be open to a pull request to add this feature

zicklag:
I'll look into this when I find the time. Do you mind assigning this to me so I don't forget?

PR #251, opened 2021-02-03
Add Exec struct for easier manipulation of exec instances

wojciechkepka:
## What did you implement: 
This PR adds possibility for a user of this crate to manipulate Exec instances through newly added Exec struct. Main motivation was to be able to check the exit_code of a given exec after its completion or to check its status while it's still running (inspect it). It also allows users to create execs without the need to start them immediately.

Any suggestions on how to handle it better, or to change naming are welcome.

Closes: #219

Briefly tested locally on one or two instances.

A new Exec struct was introduced that allows dynamically creating exec instances, manipulating them and inspecting their status.

### DIFF
--- a/examples/execinspect.rs
+++ b/examples/execinspect.rs
@@ -20,9 +20,7 @@ async fn main() {
         .build();
 
     let mut exec_id = String::new();
-    if let Some(_) = container.exec_with_id(&opts, &mut exec_id).next().await {
-        //handle output...
-    }
+    container.exec_with_id(&opts, &mut exec_id).next().await;
 
     // Inspect exec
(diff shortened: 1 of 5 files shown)


wojciechkepka:
I was also thinking of adding a wrapper struct Exec as its handled with other api endpoints if that would be something worth looking into.

wojciechkepka:
So I went ahead and added Exec as a struct and I think this solution is much cleaner than my initial attempt of passing a mutable string to a function to get an ID back.

 I've played around a little bit locally and did not notice any problems so far. Both examples worked in runtime.

PR #255, opened 2021-02-06
Api cleanup

wojciechkepka:
## What did you implement:
Thats another shot at cleaning up the code. I added helper traits JsonOpts and QueryOpts that allow for unified serialization of almost all options. The options now hold the keys as a `String` instead of `&'static str` which allows adding a Deserialize impl to them.

I also went ahead and made most of the functions have a generic interface (wherever it made sense) so now it's for example easier to pass an arbitraty iterator that produces environment variables to BuildOptionsBuilder::env because it takes an `IntoIterator<AsRef<str>> + Serialize`.

## What (if anything) should be mentioned in the CHANGELOG:

There might be some tweaking needed with arguments passed to functions but in general most of them got a generic interface which should accept more options.

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -617,28 +617,39 @@ impl ContainerOptionsBuilder {
         ContainerOptionsBuilder { name: None, params }
     }
 
-    pub fn name(
+    pub fn name<S>(
         &mut self,
-        name: &str,
-    ) -> &mut Self {
-        self.name = Some(name.to_owned());
+        name: S,
+    ) -> &mut Self
+    where
+        S: Into<String>,
+    {
+        self.name = Some(name.into());
         self
     }
 
     /// Specify the working dir (corresponds to the `-w` docker cli argument)
-    pub fn working_dir(
+    pub fn working_dir<S>(
         &mut self,
-        working_dir: &str,
-    ) -> &mut Self {
-        self.params.insert("WorkingDir", json!(working_dir));
+        working_dir: S,
+    ) -> &mut Self
+    where
+        S: AsRef<str>,
+    {
+        self.params
+            .insert("WorkingDir", json!(working_dir.as_ref()));
         self
     }
 
     /// Specify any bind mounts, taking the form of `/some/host/path:/some/container/path`
-    pub fn volumes(
+    pub fn volumes<I, S>(
         &mut self,
-        volumes: Vec<&str>,
-    ) -> &mut Self {
+        volumes: I,
+    ) -> &mut Self
+    where
+        I: IntoIterator<Item = S> + Serialize,
+        S: AsRef<str>,
+    {
         self.params.insert("HostConfig.Binds", json!(volumes));
         self
     }
@@ -650,12 +661,15 @@ impl ContainerOptionsBuilder {
         self
     }
 
-    pub fn expose(
+    pub fn expose<S>(
         &mut self,
         srcport: u32,
-        protocol: &str,
+        protocol: S,
         hostport: u32,
-    ) -> &mut Self {
+    ) -> &mut Self
+    where
+        S: AsRef<str>,
+    {
         let mut exposedport: HashMap<String, String> = HashMap::new();
         exposedport.insert("HostPort".to_string(), hostport.to_string());
 
@@ -674,7 +688,7 @@ impl ContainerOptionsBuilder {
             port_bindings.insert(key.to_string(), json!(val));
         }
         port_bindings.insert(
-            format!("{}/{}", srcport, protocol),
+            format!("{}/{}", srcport, protocol.as_ref()),
             json!(vec![exposedport]),
         );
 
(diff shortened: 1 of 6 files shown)


wojciechkepka:
I also added a `api_doc!` macro that automatically generates a hyperlink to current version of docker reference api.

elihunter173:
What's the reason to want to deserialize things like TagOptions? Adding `.into()` on every string literal and requiring that all of those strings be on the heap seems like an unnecessary hit for what I think should be a really uncommon use case.

Big :+1: from me for `api_doc!()` tho

src/builder.rs
```diff
@@ -1,18 +1,46 @@
...
-    collections::{BTreeMap, HashMap},
+    collections::HashMap,
     hash::Hash,
     iter::{IntoIterator, Peekable},
     time::Duration,
 };
 use url::form_urlencoded;
 
-#[derive(Clone, Serialize, Debug)]
+pub type Params<T> = HashMap<String, T>;
```
elihunter173, line 15: See my earlier comment.

src/lib.rs
```diff
@@ -63,20 +63,71 @@ use serde_json::Value;
...
+    () => {
+        "https://docs.docker.com/engine/api"
+    };
+}
+
+macro_rules! API_VERSION {
+    () => {
+        "v1.41"
+    };
+}
```
elihunter173, line 76: These should be consts.
```suggestion
const API_REFERENCE_URL: &'static str = "https://docs.docker.com/engine/api";
const API_VERSION: &'static str = "v1.41";
```

src/lib.rs
```diff
@@ -63,20 +63,71 @@ use serde_json::Value;
...
 pub type Result<T> = std::result::Result<T, Error>;
 
-/// Entrypoint interface for communicating with docker daemon
+api_doc! {
+"Entrypoint interface for communicating with docker daemon", =>
 #[derive(Clone)]
 pub struct Docker {
     transport: Transport,
-}
+}}
```
elihunter173, line 123: Hmmm... Now that I'm looking at the code and usage of `api_doc!()` I'm less of a fan. It's a lot of magic and not very readable (it even breaks rustfmt) for a pretty small gain.

elihunter173 (changes requested):
I think making the API consistently use Into<String> is good for usage. Again, I don't see why Params should use a `String` key because I can't imagine why these structs should ever be deserialized. Also, `api_doc!()` sounded good in theory but I think it makes the code unnecessarily hard to read.

I'd like to review again if those two changes get reverted.

src/lib.rs
wojciechkepka: Yeah, after second thought I think you're right that it's not worth the added complexity, although having a direct link to API in docs is actually a nive quality of life improvement. 

I went ahead and created #259 for documenting api endpoints, maybe we can come up with another way of solving this.

src/lib.rs
wojciechkepka: I couldn't make them consts because concat! macro that works on compile time only works with string literals. I've tried using consts but no luck.

wojciechkepka:
I've left only the generic parameters changes, and got rid of api_doc! and left builder.rs untouched for now. 

If you don't find anything wrong with this then I guess it will be ready to merge.

elihunter173 (commented):
@softprops should have the final say, but these changes look good to me. I think for consistency all endpoints should take `AsRef<str>` rather than some taking `Intro<String>`

src/builder.rs
```diff
@@ -617,28 +617,39 @@ impl ContainerOptionsBuilder {
...
-    pub fn name(
+    pub fn name<S>(
         &mut self,
-        name: &str,
-    ) -> &mut Self {
-        self.name = Some(name.to_owned());
+        name: S,
+    ) -> &mut Self
+    where
+        S: Into<String>,
```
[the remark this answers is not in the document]
wojciechkepka: Well this method is different than others because the `name` it takes is actually later used as a `String` on contrary to other methods that take a `AsRef<str>` and us this argument as `&str`. So I think it actually makes more sense like this. But if i didn't convince you then sure enough I can make all of them `AsRef<str>` and then later call `to_string`

src/builder.rs
```diff
@@ -1014,49 +1077,39 @@ impl ExecContainerOptions {
...
-            self.params
-                .entry("Cmd")
-                .or_insert_with(Vec::new)
-                .push(cmd.to_owned());
-        }
+        cmds: I,
+    ) -> &mut Self
+    where
+        I: IntoIterator<Item = S> + Serialize,
+        S: Into<String>,
```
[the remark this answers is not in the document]
wojciechkepka: this one actually should be `AsRef<str>`

src/builder.rs
```diff
@@ -1382,18 +1440,24 @@ impl ImageListOptionsBuilder {
...
-    pub fn filter_name(
+    pub fn filter_name<S>(
         &mut self,
-        name: &str,
-    ) -> &mut Self {
-        self.params.insert("filter", name.to_owned());
+        name: S,
+    ) -> &mut Self
+    where
+        S: Into<String>,
```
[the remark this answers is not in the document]
wojciechkepka: this one expects a String so I can call to_string

wojciechkepka:
What do you think about requiring `Into<String>` rather than `AsRef<str>`? All of them are used As string so I think it's more idomatic to call `.into()` rather than `.as_ref().to_string()`?

wojciechkepka:
On second thought perhaps it's better to have a unified interface. Ready for review.

Issue #241, opened 2020-12-04
Won't compile on Windows on master branch


arthmis:
- Crate version:
I'm using 17651193edad67f10f9e139d5b86b2d31665c0a6 of the master branch to get support for latest version of futures.
- OS: I'm on Windows 10
- Output of running `docker version` on the command line:
```
Client: Docker Engine - Community
 Cloud integration: 1.0.2
 Version:           19.03.13
 API version:       1.40
 Go version:        go1.13.15
 Git commit:        4484c46d9d
 Built:             Wed Sep 16 17:00:27 2020
 OS/Arch:           windows/amd64
 Experimental:      false

Server: Docker Engine - Community
 Engine:
  Version:          19.03.13
  API version:      1.40 (minimum version 1.12)
  Go version:       go1.13.15
  Git commit:       4484c46d9d
  Built:            Wed Sep 16 17:07:04 2020
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          v1.3.7
  GitCommit:        8fba4e9a7d01810a393d5d25a3621dc101981175
 runc:
  Version:          1.0.0-rc10
  GitCommit:        dc9208a3303feef5b3839f4323d9beb36df0a9dd
 docker-init:
  Version:          0.18.0
  GitCommit:        fec3683
```
The error I'm getting is:
```
thread 'main' panicked at 'Unix socket support is disabled', C:\Users\Buttah\.cargo\git\checkouts\shiplift-48fc668b583d42d4\1765119\src\lib.rs:975:21
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace     
error: process didn't exit successfully: `target\debug\amazon-price-scraper.exe` (exit code: 101)
```
My Cargo.toml is:
```toml
[package]
name = "some-name"
version = "0.1.0"
authors = []
edition = "2018"

[dependencies]
scraper = "0.12.0"
prettytable-rs = "0.8.0"
textwrap = "0.12.1"
url = "2.2.0"
anyhow = "1"
async-std = {version = "1", features = ["attributes", "tokio03", "tokio02"]}
futures = "0.3.8"
cdrs-tokio = "1"
chrono = "0.4"
reqwest = { version = "0.10.9", features = ["brotli", "gzip", "native-tls", "cookies"] }
shiplift = { git = "https://github.com/softprops/shiplift.git", default-features = false, features = ["tls"] }
```

[...]

PR #256, opened 2021-02-06
Explicit Initialization

wojciechkepka:
## What did you implement:
A safe way to initialize `Docker`. Current initialization methods can panic when uri is malformed. This PR adds a new safe way to initialize both TCP and UNIX clients. There are also multiple unwraps that I got rid of .

Running tests locally.

This is a big breaking change. It completely overhauls initialization of Docker. There are now 3 ways to initialize Docker:
 1. `Docker::new(uri)` which takes a `Into<String>` and chooses appropriate transport based on the scheme. This method errors out if uri is malformed.
 2. `Docker::from_env()` works as old `Docker::new` and reads DOCKER_HOST environment variable for the uri. What changed is that it now returns a Result.
 3. `Docker::default()` this is a way to initialize Docker without a Result. It defaults to `unix:///var/run/docker.sock`.


src/lib.rs
```diff
@@ -1021,59 +1023,95 @@ fn get_docker_for_tcp(tcp_host_str: String) -> Docker {
...
-            None => panic!("Unix socket support is disabled"),
+            None => Err(Error::UnsupportedSocket { socket: "unix" }),
         }
     }
 
     /// Creates a new docker instance for a docker host
     /// listening on a given Unix socket.
     #[cfg(feature = "unix-socket")]
-    pub fn unix<S>(socket_path: S) -> Docker
+    fn unix<S>(socket_path: S) -> Docker
```
elihunter173, line 1114: Why was this made private?

src/lib.rs
```diff
@@ -1087,29 +1125,11 @@ impl Docker {
...
-                    path: host.path().to_owned(),
-                },
-            },
-
-            #[cfg(not(feature = "unix-socket"))]
-            Some("unix") => panic!("Unix socket support is disabled"),
-
-            _ => get_docker_for_tcp(tcp_host_str),
-        }
+    fn tcp<S>(socket_path: S) -> Result<Docker>
```
elihunter173, line 1128: Same here. Why was this made private?

elihunter173 (commented):
Big fan of these changes overall. Confused why `tcp` and `unix` were made private tho.

src/lib.rs
```diff
@@ -1021,59 +1023,95 @@ fn get_docker_for_tcp(tcp_host_str: String) -> Docker {
...
+
+        if let Some(scheme) = uri.scheme() {
+            match scheme.as_str() {
+                "tcp" | "http" | "https" => Docker::tcp(uri_str),
+                s => Err(Error::UnsupportedScheme { scheme: s.into() }),
+            }
+        } else {
+            Err(Error::UnsupportedScheme { scheme: "".into() })
+        }
+    }
```
thomaseizinger, line 1097: Given you are a fan of `match` vs `if let`, what do you think of https://github.com/softprops/shiplift/commit/879a5895496edf1b61cdec22d4404e5c0109f14a?

thomaseizinger (commented):
Liking this as well!

src/lib.rs
```diff
@@ -1021,59 +1023,95 @@ fn get_docker_for_tcp(tcp_host_str: String) -> Docker {
...
+
     /// constructs a new Docker instance for a docker host listening at a url specified by an env var `DOCKER_HOST`,
-    /// falling back on unix:///var/run/docker.sock
-    pub fn new() -> Docker {
+    /// falling back to [DEFAULT_SOCKET_PATH](DEFAULT_SOCKET_PATH))
+    pub fn from_env() -> Result<Docker> {
```
softprops, line 1101: I think there's a non breaking approach to the change you are looking for. A common approach to providing alternative methods that return Result is to provide an alternative try_xxx method. Can we keep new and the Default impl as is and provide a try_new() which returns a result? That would give you what you'd like, avoid more breaking api changes and possibly be a bit more idiomatic

softprops (changes requested):

src/lib.rs
softprops: I'd also like to keep these public for cases users would like a more explicit constructor

wojciechkepka: Wasn't sure about this, but I agree keeping them public is a better choice.

src/lib.rs
wojciechkepka: That looks awesome! Will definitely use it.

src/lib.rs
wojciechkepka: I went ahead with your suggestion and renamed new -> try_new and brought back old behavior. I also added a big warning on the new method that it can panic in some cases and added a note how to safely initialize an instance.

wojciechkepka: I decided to keep examples as from_env() to promote a safer way of initialization/

wojciechkepka: And also renamed from_env -> try_env to be consistent.

src/lib.rs
```diff
@@ -1055,7 +1055,15 @@ where
 }
 
 impl Default for Docker {
-    /// Returns a Docker instance connected to a unix socket at [DEFAULT_SOCKET_PATH](DEFAULT_SOCKET_PATH)
+    /// Constructs a new Docker instance for a docker host listening at a url specified by an env var `DOCKER_HOST`,
+    /// falling back to unix socket with path [DEFAULT_SOCKET_PATH](DEFAULT_SOCKET_PATH)) if
+    /// `unix-socket` feature is enabled.
+    ///
+    /// WARNING!
```
thomaseizinger, line 1055: If you want to be more [rustdoc-idiomatic](https://doc.rust-lang.org/book/ch14-02-publishing-to-crates-io.html#commonly-used-sections), then this section could have a header `# Panics`.

src/lib.rs
```diff
@@ -1110,8 +1110,12 @@ impl Docker {
 
     /// Creates a new docker instance for a docker host
     /// listening on a given Unix socket.
+    ///
+    /// socket_path is a part of uri that comes after the scheme.
+    ///
+    /// For example a uri `unix:///run/docker.socket` has a socket_path == "/run/docker.sock"
```
thomaseizinger, line 1134: Minor typo:

```suggestion
    /// For example a uri `unix:///run/docker.sock` has a socket_path == "/run/docker.sock"
```

thomaseizinger (commented):
Two more drive-by comments :)

src/lib.rs
```diff
@@ -1021,57 +1019,121 @@ fn get_docker_for_tcp(tcp_host_str: String) -> Docker {
...
+            None => Err(Error::UnsupportedScheme { scheme: "".into() }),
         }
     }
 
-    /// Creates a new docker instance for a docker host
-    /// listening on a given Unix socket.
+    /// Constructs a new Docker instance for a docker host listening at a url specified by an env var `DOCKER_HOST`,
+    /// falling back to unix socket with path [DEFAULT_SOCKET_PATH](DEFAULT_SOCKET_PATH)) if
+    /// `unix-socket` feature is enabled, otherwise returns an error.
+    pub fn try_env() -> Result<Docker> {
```
elihunter173, line 1122: I think `try_from_env()` is more obvious name. `try_env()` sounds like it's taking some action to me

wojciechkepka: Agreed and done.

src/lib.rs
```diff
@@ -1161,57 +1163,121 @@ fn get_docker_for_tcp(tcp_host_str: String) -> Docker {
...
+impl Default for Docker {
+    /// Constructs a new Docker instance for a docker host listening at a url specified by an env var [DOCKER_ENV_VAR](DOCKER_ENV_VAR),
+    /// falling back to unix socket with path [DEFAULT_SOCKET_PATH](DEFAULT_SOCKET_PATH)) if
+    /// `unix-socket` feature is enabled.
+    ///
+    /// # Panics
+    ///
+    /// This function panics when a given url is incorrect or if the url specified is a unix socket
+    /// and `unix-socket` feature is disabled. To safely initialize Docker instance from
+    /// environment variable use [Docker::try_from_env](Docker::try_from_env] or
```
matthiasbeyer, line 1203: Messed up URL link syntax :laughing:

src/lib.rs
```diff
@@ -1161,57 +1163,121 @@ fn get_docker_for_tcp(tcp_host_str: String) -> Docker {
...
-    /// falling back on unix:///var/run/docker.sock
+    /// Constructs a new Docker instance for a docker host listening at a url specified by an env var [DOCKER_ENV_VAR](DOCKER_ENV_VAR),
+    /// falling back to unix socket with path [DEFAULT_SOCKET_PATH](DEFAULT_SOCKET_PATH)) if
+    /// `unix-socket` feature is enabled.
+    ///
+    /// # Panics
+    ///
+    /// This function panics when a given url is incorrect or if the url specified is a unix socket
+    /// and `unix-socket` feature is disabled. To safely initialize Docker instance from
+    /// environment variable use [Docker::try_from_env](Docker::try_from_env] or
```
matthiasbeyer, line 1220: Here as well.

matthiasbeyer (changes requested):
I did not do an in-depth review here, just a very basic one.

Either way, I'm all in for a big rebase before this is merged, because the commit history looks not so clean :smile:

wojciechkepka:
I rebased to current master, cleaned up the history and fixed the links as you suggested :)

wojciechkepka:
I completely agree, libraries shouldn't access environment variables and the `get_docker_for_tcp` also shouldn't use `DOCKER_CERT_PATH` as well as `DOCKER_TLS_VERIFY` but perhaps take this as arguments to the `Docket::tcp` constructor.

I would perhaps get rid of the `new`, `try_from_env` and `Default` and rename `try_new` to `new`. I think it's worth keeping a `new` method that actually handles features and parsing uri properly, but if we also add new parameters to `Docker::tcp` then perhaps the `new` method might be pointless.

wojciechkepka:
I think Rust is a language where explicitness is a key value thus I went ahead and rebased this PR to new master and applied initial changes mentioned in my previous comment.

What do you think of this approach?

thomaseizinger:
Like I mentioned [here](https://github.com/softprops/shiplift/issues/241#issuecomment-792362232), I am in favor of a "smart" initialization function that makes it easier for people to use shiplift in a cross-platform context.

However, I also think that the changes in this branch provide a much better foundation for achieving this. I think it would be good to strive for backwards-compatible changes so perhaps there is a middle ground here where we can retain the public API as much as possible but still change things under the hood to already be organized into the different ways of initializing a docker client.

wojciechkepka:
I think there should be no need to stay backwards compatible while the public API is still pre-1.0 but that's just my 2 cents.

matthiasbeyer:
> Perhaps we might want to verify the validity of the passed uri and return Results?

IMO yes. If the user has a `Docker` object, they should be able to use it because initialization worked. If they can create a `Docker` that does not work, that's unintuitive, IMHO.

wojciechkepka:
I pushed some changes, now tcp, tls and Unix constructors take only the host path that comes after the Scheme as an argument and the new method does the magic matching based on features and the scheme. 

I've decided to keep tls as a separate constructor as it complicates some things.

I'm not entirely sure how to validate the host path without connecting to it to verify if the connector works.

wojciechkepka:
So I cleaned up the commit history and code a bit, any suggestions? Is this the right direction?

src/docker.rs
```diff
@@ -107,29 +109,78 @@ fn get_docker_for_tcp(tcp_host_str: String) -> Docker {
...
+    /// on provided `uri`.
+    ///
+    /// Supported schemes are:
+    ///  - `unix://` with feature `unix-socket` enabled, otherwise returns an Error
+    ///  - `tcp://`
+    ///  - `http://`
+    ///
+    ///  To create a Docker instance utilizing TLS use explicit [Docker::tls](Docker::tls)
+    ///  constructor.
+    pub fn new<S>(uri: S) -> Result<Docker>
```
matthiasbeyer, line 149: I am not sure whether I like the automatic finding of the scheme here, because a user must rely on the correct implementation of the parsing.
Maybe we can have something like this:

```
Docker::new(DockerConnectionScheme::from_str("unix://...")?)
```

So that we keep this general idea, but also give the user the option to be explicit, be having him construct a `DockerConnectionScheme` themselves:

```
pub enum DockerConnectionScheme {
    UnixSocket(PathBuf),
    Tcp(url::Url), // or similar - strongly typed
    Http(url::Url), //...
}
```

And with an appropriate `std::str::FromStr` implementation that does the parsing. This way a user can chose what they want!

What do you think?

---

(Above code is just a quick mockup of my idea, details may be different)

matthiasbeyer (changes requested):
I like the direction this is going, I'm just not yet 100% happy with the `Docker::new()` interface :laughing: 

I hope you don't mind...

wojciechkepka:
Sorry for the hustle, I'm actually working on my own library now so I'm not sure I'll be able to contribute here too. I'll close this issue for now.

matthiasbeyer:
Would you mind if I take your commits and work with them? From this PR as well as from the other PRs you closed?

wojciechkepka:
Sure, not a problem. Feel free to rebase the PRs and update them accordingly :)

PR #258, opened 2021-02-07
Add Image(s)::save() and deprecate export()

### LINKED ISSUE #184: Question: should Image.export be renamed to save?
Should [`export`](https://docs.rs/shiplift/0.5.0/shiplift/struct.Image.html#method.export) be called `save` instead? Just to be consistent with cmdline `docker` and their Go API ([ImageSave](https://godoc.org/github.com/docker/docker/client#Client.ImageSave)).

This is a very minor nit because `export`'s documentation and the example do illustrate that it saves things to a tarball. I don't want to bike shed too much, but the name kind of threw me off when I first looked at it.

elihunter173:
## What did you implement:

I renamed `Image::export()` to `Image::save()` and did the same for `Images` to make their naming inline with the [Go API](https://pkg.go.dev/github.com/docker/docker/client#Client.ImageSave) and the docker CLI.

I discovered an old issue #184 that mentions this and I tend to agree that it would be good to use consistent terminology.

Closes: #184 

Ran tests locally.

Mention that `Image(s)::export()` has been deprecated in favor of `Image(s)::save()`. (Already done.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 * support for uploading tar to container [#239](https://github.com/softprops/shiplift/pull/239)
 * fix registry authentication to use URL-safe base64 encoding [#245](https://github.com/softprops/shiplift/pull/245)
 * add StopSignal and StopTimeout to ContainerOptionsBuilder [#248](https://github.com/softprops/shiplift/pull/248)
+* add Image(s)::save() and deprecate Image(s)::export() [#257](https://github.com/softprops/shiplift/pull/257)
 
 # 0.6.0
 
--- a/examples/save.rs
+++ b/examples/save.rs
@@ -8,7 +8,7 @@ async fn main() {
     let docker = Docker::new();
     let id = env::args().nth(1).expect("You need to specify an image id");
 
-    let mut export_file = OpenOptions::new()
+    let mut file = OpenOptions::new()
         .write(true)
         .create(true)
         .open(format!("{}.tar", &id))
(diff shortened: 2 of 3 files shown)


elihunter173:
Just realized that I made a typo in one of the return types. Should compile and pass all tests again.

vv9k (approved):

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@
 * support for uploading tar to container [#239](https://github.com/softprops/shiplift/pull/239)
 * fix registry authentication to use URL-safe base64 encoding [#245](https://github.com/softprops/shiplift/pull/245)
 * add StopSignal and StopTimeout to ContainerOptionsBuilder [#248](https://github.com/softprops/shiplift/pull/248)
+* add Image(s)::save() and deprecate Image(s)::export() [#257](https://github.com/softprops/shiplift/pull/257)
```
matthiasbeyer, line 18: I guess this has to be in the changelog section for the next release, not the existing one, right? :laughing:

matthiasbeyer, line 18: I guess this have to be in the changelog for the next one, not for the existing one, right? :laughing:

matthiasbeyer (changes requested):
Looks good in general.

Just a quick question: We're deprecating functions only on master and delete them before the release in this crate, is that correct? Because I'd imagine that we rather deprecate, release, then remove and make them gone in one release further. So the deprecation would remove them _after_ the 0.8.0 release.

I'm fine either way, just confused why it is done this way and not the other.

Issue #259, opened 2021-02-08
Add a link to Docker api reference on each api endpoint in this library


wojciechkepka:
I had some initial attempts at adding hyperlinks to Docker API Reference in #255 but we came to the agreement that using a macro like my proposed `api_doc!` that would wrap a struct or a function and annotate it with extra `#[doc = ""]` attributes was making the code unreadable and broke rustfmt.

Perhaps there is another approach to this problem. I see this as a nice quality of life improvement to reading the docs and I think its worth the effort to automatize it (maybe with a proc macro?)

matthiasbeyer:
Are you still planing to approach this, even though your PRs were closed unmerged? :pray:

wojciechkepka:
Well I would definitely be interested but perhaps first we could come up with a proper way to add this before I submit any changes.

elihunter173:
I think I mentioned it earlier, but I think a line like this at the end of doc comments would be sufficient since the API links are unlikely to change and even if they do a simple sed / search and replace should be able to find them all pretty easily.

```rust
/// [API Reference](api-link)
```

PR #260, opened 2021-02-08
Document API endpoints

wojciechkepka:
## What did you implement:
`api_doc` procedural macro that can be used to easily annotate functions and methods to automatically add hyperlinks to docker api reference website. Before I go on and finish documenting all endpoints I wanted to ask for a review and potential suggestions.

I think a proc macro is the best solution here. We have to compile syn and quote anyways as they are in the dependency tree so it's not that costly.

Closes: #259

Ran `cargo doc --no-deps` and got expected results

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -37,6 +37,7 @@ serde_json = "1.0"
 tar = "0.4"
 tokio = "1.0"
 url = "2.1"
+shiplift-macro = { version = "0.1", path = "./shiplift-macro" }
 
 [dev-dependencies]
 env_logger = "0.8"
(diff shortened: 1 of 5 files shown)


elihunter173 (commented):
This seems like a lot of code to me to add `[API Reference](api-link)` to every method. I understand and appreciate the reasoning, but I can't help but feel like it would be easier to just manually put this at the end of every methods doc comment. If we need to update the links in the future, sed or even find and replace should be able to change it really easily.

If this was like the initial ~5 lines of code I had hoped for, I'd be all for this but as it stands ~180 lines of non-trivial code seems like unnecessary overhead on maintainers.

PR #264, opened 2021-02-10
Add missing fields to `ImageDetails` and `History`

saruman9:
## What did you implement:

Sync with API version 1.41.

Compare with Docker CLI commands (`inspect` and `history`).

Some fields are now optional (`Option<T>`).


src/rep.rs
```diff
@@ -460,8 +508,12 @@ pub struct History {
     #[serde(deserialize_with = "datetime_from_unix_timestamp")]
     pub created: DateTime<Utc>,
     #[cfg(not(feature = "chrono"))]
-    pub created: u64,
+    pub created: i64,
     pub created_by: String,
+    // Not Required
+    pub tags: Option<Vec<String>>,
+    pub size: i64,
+    pub comment: String,
```
elihunter173, line 516: I think the "Not Required" comment should be removed. If it's just about the `tags` field then I think that `Option` conveys that but if its for `size` or `comment` then it's confusing because they're not `Option<T>`

elihunter173 (commented):
Overall looks good. Could you please update the CHANGELOG and specifically call out what fields have been made `Option<T>`? The only one I notice is `ImageDetails::config`

saruman9:
> Could you please update the CHANGELOG and specifically call out what fields have been made `Option<T>`?

I updated, could you please check.

src/rep.rs
saruman9: In [Docker doc](https://docs.docker.com/engine/api/v1.41/#operation/ImageHistory) `Tags` field are required, but it's not true (I checked it locally). I wanted to point this out so that other developers don't think this is a mistake. Do you think it is not necessary?

wojciechkepka: I don't think pointing it out makes sense as this is the case for a lot of fields on most of the structs here. The api is very loose on saying what actually is Nullable and what's not. 

When I added missing fields to ContainerDetails I basically made all fields required where it said and then I inspected a container without any filters and options and ensured it was a container that has no volumes/networks etc. After that it is trial and error with serde deserialization, and adjusting fields to options where it errors out.

saruman9: > The api is very loose on saying what actually is Nullable and what's not.

Agree with you.

>     // Not Required
Removed it.

src/rep.rs
```diff
@@ -194,29 +201,35 @@ pub struct HostConfig {
...
-pub struct Config {
-    pub attach_stderr: bool,
-    pub attach_stdin: bool,
-    pub attach_stdout: bool,
+pub struct ContainerConfig {
+    pub hostname: Option<String>,
+    pub domainname: Option<String>,
+    pub user: Option<String>,
+    pub attach_stdin: Option<bool>,
+    pub attach_stdout: Option<bool>,
```
wojciechkepka, line 209: I don't think those fields should be `Option`al. They seem to have default values and when the value is missing the just return either `""` or `false`.

src/rep.rs
```diff
@@ -194,29 +201,35 @@ pub struct HostConfig {
...
+    pub attach_stderr: Option<bool>,
+    pub exposed_ports: Option<HashMap<String, HashMap<String, String>>>,
+    pub tty: Option<bool>,
+    pub open_stdin: Option<bool>,
+    pub stdin_once: Option<bool>,
```
wojciechkepka, line 214: Same here.

src/rep.rs
```diff
@@ -194,29 +201,35 @@ pub struct HostConfig {
...
+    pub env: Option<Vec<String>>,
     pub cmd: Option<Vec<String>>,
-    pub domainname: String,
+    pub healtcheck: Option<HealthConfig>,
+    pub args_escaped: Option<bool>,
+    pub image: Option<String>,
```
wojciechkepka, line 219: Same.

src/rep.rs
```diff
@@ -194,29 +201,35 @@ pub struct HostConfig {
...
+    pub volumes: Option<HashMap<String, HashMap<String, String>>>,
+    pub working_dir: Option<String>,
```
wojciechkepka, line 221: Same.

src/rep.rs
```diff
@@ -194,29 +201,35 @@ pub struct HostConfig {
...
     pub entrypoint: Option<Vec<String>>,
-    pub env: Option<Vec<String>>,
-    pub exposed_ports: Option<HashMap<String, HashMap<String, String>>>,
-    pub hostname: String,
-    pub image: String,
-    pub labels: Option<HashMap<String, String>>,
-    // pub MacAddress: String,
+    pub network_disabled: Option<bool>,
+    pub mac_address: Option<String>,
```
wojciechkepka, line 224: Same.

src/rep.rs
```diff
@@ -194,29 +201,35 @@ pub struct HostConfig {
...
     pub on_build: Option<Vec<String>>,
-    // pub NetworkDisabled: bool,
-    pub open_stdin: bool,
-    pub stdin_once: bool,
-    pub tty: bool,
-    pub user: String,
-    pub working_dir: String,
+    pub labels: Option<HashMap<String, String>>,
+    pub stop_signal: Option<String>,
+    pub stop_timeout: Option<u64>,
```
wojciechkepka, line 228: Same.

src/rep.rs
```diff
@@ -460,8 +508,11 @@ pub struct History {
     #[serde(deserialize_with = "datetime_from_unix_timestamp")]
     pub created: DateTime<Utc>,
     #[cfg(not(feature = "chrono"))]
-    pub created: u64,
+    pub created: i64,
```
wojciechkepka, line 511: Looking at the codebase there are 85 occurances of `u64` and only 3 of `i64` so I'd stick with that.

src/rep.rs
```diff
@@ -460,8 +508,11 @@ pub struct History {
...
     pub created_by: String,
+    pub tags: Option<Vec<String>>,
+    pub size: i64,
```
wojciechkepka, line 514: Same here, `i64` -> `u64`

wojciechkepka (commented):
I'd encourage you to try out my method pointed out in another comment.

src/rep.rs
wojciechkepka: Without `volumes`

src/rep.rs
saruman9: I don't know about other fields, but for example `args_escaped` doesn't exist, when the value is missing (I checked it) and as a result - we get `SerdeJsonError`.

Do you know, where we can find information about it? Are you sure about other fields?

src/rep.rs
saruman9: I was guided by the [API documentation](https://docs.docker.com/engine/api/v1.41/#operation/ImageHistory): `integer <int64>`.

If we look at the codebase we can find other flaws as well: missing fields, don't respecting required/optional fields. I think it is better to write the code according to API documentation.

src/rep.rs
wojciechkepka: Perhaps `args_escaped` might not exist. I wanted to point out here that you should double check before making everything optional. If you're getting a `SerdeJsonError` then adjust the code accordingly and make it optional at this field.

As I pointed out in another comment the docs are very loose and the only way I know of is trial and error by first making all fields non-optional and later adjusting them one by one where it errors out. I know it's not really pleasant to do but I don't know any other way and it would be a shame to make everything an Option as accessing the fields will become quite cumbersome.

wojciechkepka: The most important fields to check I'd say are numbers, bools and strings as those tend to have a default value most of the time. So I'd go for no Options for those and Options for Vecs/Hashmaps (but that should also be checked beforehand).

saruman9: > later adjusting them one by one where it errors out

For some images these fields are, for some they are not. Where can we find so many different images (for example, `args_escaped` may be `true`, `false`, missing for different images)?

>  it would be a shame to make everything an Option as accessing the fields will become quite cumbersome

I agree with you, but it will be much worse if the user gets `SerdeJsonError` on some image that I have not tested.

src/rep.rs
wojciechkepka: Well, agreed. Perhaps it's better to go the other way and start being more strict about the type.

wojciechkepka: So I'm just trying to make sure to perhaps cleanup my PRs. Do you think that by `integer` they mean a `u64` and by `integer <int64>` it means `i64`?

src/rep.rs
wojciechkepka: I think there might be a way to test it if we use image `scratch`, for example:
 - create a directory `test` with a `Dockerfile` in it
 - The dockerfile has contents:
 ```
 FROM scratch
 ENV test test
 ```
  - build it
  - inspect it
  
This will give you the most barebone image that exists because it's `scratch` used to build images from `scratch` ;) So it should have nulls on all nullable fields.

wojciechkepka: OK, perhaps an even better solution would be to look at the `go-docker` client and check which fields have `omitempty` json option or are pointers to struct which can be null.

[Link](https://github.com/docker/go-docker/blob/master/api/types/types.go)

wojciechkepka: It looks like it actually matches the output. As you noticed `args_escaped` are optional and they have `omitempty` flag:
https://github.com/docker/go-docker/blob/4daae26030ad00e348edddff9767924ae57a3b82/api/types/container/config.go#L57

wojciechkepka: Ok, i just noticed this repo is from 2017...

wojciechkepka: If this doesn't work then I'm out of ideas on how to properly handle it and would probably be in favor of using Options all over the place unless anyone comes up with a better solution.

saruman9: Thank you for researching.

I fix some fields (bb04336).

It's strange, that some tools don't respect [official swagger API file](https://github.com/moby/moby/blob/c9bbc68e75f224c4490ff6e3121be50a612edc22/api/swagger.yaml), part of codebase generated from swagger file and part of code are wrote manually.

wojciechkepka: Awesome, looks good to me :)

CHANGELOG.md
```diff
@@ -15,6 +15,11 @@
 * support for uploading tar to container [#239](https://github.com/softprops/shiplift/pull/239)
 * fix registry authentication to use URL-safe base64 encoding [#245](https://github.com/softprops/shiplift/pull/245)
 * add StopSignal and StopTimeout to ContainerOptionsBuilder [#248](https://github.com/softprops/shiplift/pull/248)
+* make `config` field of `ImageDetails` optional [#264](https://github.com/softprops/shiplift/pull/264)
+* add `container`, `container_config`, `os_version`, `graph_driver`, `root_fs`, `metadata` fields to `ImageDetails` [#264](https://github.com/softprops/shiplift/pull/264)
+* rename `shiplift::rep::Config` to `shiplift::rep::ContainerConfig` [#264](https://github.com/softprops/shiplift/pull/264)
+* add missing fields ([API version 1.41](https://docs.docker.com/engine/api/v1.41/#operation/ImageInspect)) to `ContainerConfig` and make some fields optional [#264](https://github.com/softprops/shiplift/pull/264)
```
wojciechkepka, line 21: So from what I've looked none of the fields that were already available were made optional. So this can just say that missing fields were added,

wojciechkepka (changes requested):

CHANGELOG.md
saruman9: Yes, you are right. Fixed.

Thank you for reviewing!

wojciechkepka (approved):

Issue #202, opened 2019-11-26
missing fields in the result of inspect


jschwinger23:
as per the doc: https://docs.rs/shiplift/0.5.0/shiplift/rep/struct.HostConfig.html, there are only 12 fields supported in HostConfig, leaving bunch of fields like `CpuQuota` and `CpuPeriod` unreachable.

here's docker inspect result example for your reference: https://docs.docker.com/engine/api/v1.24/

Issue #172, opened 2019-05-22
strongly typed ContainerDetails.


danieleades:
some of the types in the ContainerDetails struct are a little hard to work with.
take ContainerDetails.NetworkSettings.ports for example-

`ports: Option<HashMap<String, Option<Vec<HashMap<String, String>>>>>`

it's not obvious what this means, or particularly ergonomic
is the outer hashmap a map from protocol to port number?
is the inner hashmap from source port to host port?
why in god's name is there a `Vec<HashMap>`?

my use case: I always use one tcp port mapping- a consistent source port, and a randomised host port. how do I extract the host port from this struct?

something like-
`ports.unwrap().get("tcp").unwrap().unwrap()[0].get(src_port).unwrap()`
(i'm genuinely asking, as well as having a whinge!)?

it's not my favourite

danieleades:
any update on this?

PR #266, opened 2021-02-10
Add missing fields to ContainerDetails

wojciechkepka:
## What did you implement:
Added missing fields to `ContainerDetails` and reordered them to resemble the same order as docker api reference.

Closes: #202
Closes: #172

Tested locally.

- `ContainerDetails` now contains all fields available in docker.
- `HostConfig.port_bindings` inner elements now have a clear type `PortBinding` instead of `HashMap<String, String>`
- `Config` is now renamed to `ContainerConfig`


wojciechkepka:
Rebased to current master, I guess ready to merge if there are no suggestions.

matthiasbeyer:
Hm. Generally this looks good IMO, but is there a way to add automated (or manual) tests for this? Like... an example that one can execute to see the data points added in the API here?

Issue #257, opened 2021-02-07
Refactor Modules by API Endpoint


elihunter173:
Currently a lot of functionality is split between `lib.rs` and `builder.rs`. I think it would make sense to split functionality into modules based off of their endpoint they hit (e.g. `container.rs` and `image.rs`) since that should make it easier for maintainers to maintain and new members to navigate the docs. We could keep most of the API somewhat unchanged through re-exports.

I plan to take this on if there's interest coming by wanted to make sure before I sent the PR in.

wojciechkepka:
I had the same thoughts recently so I'm all for this change. I must say it would make it much clearer to navigate the code base.

PR #267, opened 2021-02-10
Split lib.rs, builder.rs, and rep.rs into endpoint-specific modules

elihunter173:
**This is a non-breaking change.** Since this is a large PR, I want to ensure that no code inside of shiplift changes and no code outside of shiplift has to immediately change. To do this there ~~will be~~ are re-exports of all items inside their old module (with deprecation warnings) so that they will be accessible both in their new module and old module.

## What did you implement:

I split up `lib.rs`, `builder.rs`, and `rep.rs` into `container.rs`, `docker.rs`, `exec.rs`, `image.rs`, `network.rs`, `service.rs`, and `volume.rs` to reflect what part of the docker API the components deal with. These were modeled after the [Docker API's sections](https://docs.docker.com/engine/api/v1.41/) so that it should be intuitive for both maintainers and users to find what code or documentation to look at based off of what operation they want to do, just by reading the API. It should also nicely extend as shiplift adds support for more endpoints. (For example, it was easy to merge #263.)

I am still working on changes so I've marked this PR as a draft. Here's a short list of the things I still have left to do
* [x] Move the remaining `lib.rs` into `docker.rs`.
* [x] Include re-exports for backwards compatibility.
* [x] Add module-level documentation to each module.

By the nature of this PR, this is disruptive to other PRs. Every currently opened PR will have to be modified after this one lands and every PR that lands before this one means that this PR will have to be modified. In hindsight I wish I did this when the project was less active but alas. Luckily, this PR introduces no code changes, instead just relocating the code, so it should be very easy to merge new changes since it just involves completely moving chunks of code to different files and then changing the imports.

Closes: #257

* [x] Ensure all shiplift tests pass.
* [x] Ensure nothing in public API is removed, only deprecated.
* [x] Ensure that various public projects still build after this PR lands.
  * [x] [`elihunter173/codie-discord`](https://github.com/elihunter173/codie-discord)
  * [x] [`wojciechkepka/pkger` on `rewrite`](https://github.com/wojciechkepka/pkger/tree/rewrite) Works once #269 is also merged
  * [ ] [`aixigo/PREvant`](https://github.com/aixigo/PREvant) Couldn't figure out how to build an unchanged master :disappointed: 
  * [x] [`~matthiasbeyer/butido`](https://git.sr.ht/~matthiasbeyer/butido)
  * ~~[ ] [`aws/aws-nitro-enclaves-cli`](https://github.com/aws/aws-nitro-enclaves-cli)~~ Actually they're on 0.6.0

* `shiplift::{container, docker, exec, image, network, service, volume}` have been added as modules containing their respective items in `shiplift::builder`, `shiplift::rep`, and `shiplift`.
* `shiplift::builder` and `shiplift::rep` have been deprecated in favor the more specific `shiplift::{container, docker, exec, image, network, service, volume}` modules.

*Note:* Any `shiplift::rep` structs that had name collisions with other structs were renamed with `Info` at the end of their name. For example, `shiplift::Service` and `shiplift::rep::Service` and collide, so when they were moved to `shiplift::service::Service` and `shiplift::service::ServiceInfo` respectively.


wojciechkepka:
Love the changes. It will make contributing much more pleasant (at least for me). I dont mind rebasing my PRs if yours will get merged first.

elihunter173:
I just added backwards-compatible type aliases for the old `shiplift::builder` and `shiplift::rep` modules. This adds one new temporary dependency on [`paste`](docs.rs/paste/).

Since doing the re-exporting fully manually would be incredibly boilerplate heavy and time-consuming, I wrote a fairly small (and hopefully legible) declarative macro that would create the modules with the re-exports and the appropriate deprecation messages. You can't concatenate strings inside of attribute macros normally, which I wanted to do to recommend the new item, but `paste` enables that through proc-macro magic, so I added that dependency so that we could have high-quality deprecation messages on every item inside of the these modules (see below). 

If that dependency or all the macro isn't desirable for compile times, legibility, etc, I'm completely fine with just replacing the macro with their expanded form using `cargo-expand` since this is temporary and should never change. (This would be ~5000 lines of code vs the current ~75.)

Here's an example of the deprecation warning you'd see.

```rust
use shiplift::rep::Volume;

fn main() {
    println!("Hello, world!");
}
```

```
warning: use of deprecated type alias `shiplift::rep::Volume`: Please use `shiplift::volume::VolumeInfo`. This will be removed in 0.8.0.
 --> src/main.rs:1:5
  |
1 | use shiplift::rep::Volume;
  |     ^^^^^^^^^^^^^^^^^^^^^
  |
  = note: `#[warn(deprecated)]` on by default
```

wojciechkepka:
Fow what it's worth I just compared your branch with master in terms of compile times and I don't see any meaningful difference.
```
eli on  modules-match-endpoints is 📦 v0.7.0 via 🦀 v1.49.0
❯ cargo clean
eli on  modules-match-endpoints is 📦 v0.7.0 via 🦀 v1.49.0
❯ time bash -c 'cargo b --release >/dev/null 2>&1'

real    0m23.547s
user    2m56.392s
sys     0m7.027s
```
```
shiplift on  master is 📦 v0.7.0 via 🦀 v1.49.0
❯ cargo clean

shiplift on  master is 📦 v0.7.0 via 🦀 v1.49.0
❯ time bash -c 'cargo b --release >/dev/null 2>&1'

real    0m24.149s
user    2m55.418s
sys     0m7.129s
```

elihunter173:
That's good. I didn't think it would because paste is pretty small since we already as syn and proc_macro2 in the dependency tree and I'm sure that our build time is dominated by the ~150 dependencies

matthiasbeyer:
Lets see whether butido builds with this change: https://builds.sr.ht/~matthiasbeyer/job/450444
(git commit: https://git.sr.ht/~matthiasbeyer/butido/commit/19fad6f167d3bd90566b0c7ae2530e200d8ee5c4)

Thanks for working on this!

matthiasbeyer:
So it seems that butido does _not_ build with this PR. This might be caused by other changes since the last release, though. I didn't check.

elihunter173:
> So it seems that butido does not build with this PR. This might be caused by other changes since the last release, though. I didn't check.

This should be because I haven't merged #272 into this branch yet, since I'm not seeing any errors about undefined types or invalid imports. (Although I do get warnings about deprecated type aliases :grin:)

I'm going to work on getting this branch caught back up to master and then it should build.

elihunter173:
butido builds now

matthiasbeyer:
Awesome, you're doing good work here. At some points, we might want to rebase this branch to latest master, right?

elihunter173:
The code should be up to date with the latest master, but I agree with rebasing. My plan was to rebase/squash it down to 1 commit once you think the PR is in a good state. I personally think it is.

matthiasbeyer:
I wouldn't even go down to 1 commit. Whatever makes sense to you of course, but I see at least one "cargo fmt" commit that could be standalone, and maybe the "Split up {lib,rep,build}.rs" could be one each.

Either way, I will review as soon as I have time (most certainly today).

Cargo.toml
```diff
@@ -38,6 +38,10 @@ tar = "0.4"
 tokio = "1.0"
 url = "2.1"
 
+# XXX: This is a temporary dependency for the reexport! macro in lib.rs. Remove
+# me before 0.9.0 is released.
```
matthiasbeyer, line 42: You mean 0.8.0, don't you?

matthiasbeyer (changes requested):

Cargo.toml
matthiasbeyer: Ah, now I see... sorry.

Okay, I guess you can rebase :smile:

matthiasbeyer:
Feel free to rebase.

elihunter173:
Just rebased it. History should be much cleaner now

matthiasbeyer (approved):
Shall we wait for another review? Not sure whether anyone wants to chime in...

wojciechkepka (approved):

wojciechkepka:
I quickly glanced over and looks good to me. Can't wait for the changes to be merged ;) it will be much easier to navigate the codebase ( at least for me ).

matthiasbeyer:
:tada:

PR #269, opened 2021-02-11
Fix `entrypoint` field in ContainerOptions

### LINKED ISSUE #268: Unable to create container with correct Entrypoint
As said in the title, currently `ContainerOptionsBuilder::entrypoint` has a `&str` as a parameter although in Dockerfile it's possible to declare an Entrypoint as an array of strings making it currently impossible to create containers that have a predefined executable with a set of args at entry.

It can be seen [here](https://docs.docker.com/engine/api/v1.41/#operation/ContainerCreate) under `Entrypoint`or [here](https://docs.docker.com/engine/reference/builder/#entrypoint)

- Crate version: 0.7.0
- Output of running `docker version` on the command line:
```
Client:
 Version:           20.10.3
 API version:       1.41
 Go version:        go1.15.7
 Git commit:        48d30b5b32
 Built:             Tue Feb  2 19:54:22 2021
 OS/Arch:           linux/amd64
 Context:           default
 Experimental:      true
 ```

vv9k:
## What did you implement:
Fixed `ContainerOptionsBuilder::entrypoint` to correctly take in an array of strings.

Closes: #268 

Verified on local setup.

- `ContainerOptionsBuilder::entrypoint` now takes an `IntoIterator<Item = AsRef<str>>` instead of `&str` as `entrypoint` parameter.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 * support for uploading tar to container [#239](https://github.com/softprops/shiplift/pull/239)
 * fix registry authentication to use URL-safe base64 encoding [#245](https://github.com/softprops/shiplift/pull/245)
 * add StopSignal and StopTimeout to ContainerOptionsBuilder [#248](https://github.com/softprops/shiplift/pull/248)
+* `ContainerOptionsBuilder::entrypoint` now correctly takes an `IntoIterator<Item = AsRef<str>>` instead of `&str` [#269](https://github.com/softprops/shiplift/pull/269)
 
 # 0.6.0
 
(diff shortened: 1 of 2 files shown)


elihunter173 (approved):

DCNick3:
Can a release including this change be made please?

UPD: this doesn't seem to be in master anymore too...

vv9k:
This crate is rather unmaintained I'd suggest looking for an alternative. I have an active fork with a few changes, shameless plug: https://github.com/vv9k/docker-api-rs

You can also checkout bollard: https://github.com/fussybeaver/bollard

DCNick3:
Oh, thanks for the heads up, I'll look into them

Issue #271, opened 2021-02-17
Liftimes for various methods incorrect


elihunter173:
Various methods have subtly incorrect lifetimes, which are hard to notice due to lifetime elision. In particular, I noticed that `Containers::get()` returns a `Container` with its lifetime bound to the `&self` reference rather than the `Docker` reference that `self` contains. I believe that is why I've experienced various "temporary does not exist long enough" errors in using shiplift. It's late and I don't concrete piece of code to show that this is a bug, but I will update this later to show that this is a bug.

Edit: I didn't even have to look far. Look at this `example/imagebuild.rs`.
https://github.com/softprops/shiplift/blob/5af822c5691a772cad36bd9a69e94419b03d4511/examples/imagebuild.rs#L6-L22

We're creating a local `images` because the lifetime of `stream` is tied to the reference to `images` in `images.build()`. If we don't assign to a local then the compiler creates a temporary `images` for `docker.images().build()` that only lives for that statement. That means that `stream` only lives for that expression and thus doesn't live long enough to reach the while loop, making the code invalid.

The rust compiler does a better job explaining...

```
error[E0716]: temporary value dropped while borrowed
  --> src/main.rs:12:22
   |
12 |     let mut stream = docker.images().build(&options);
   |                      ^^^^^^^^^^^^^^^                - temporary value is freed at the end of this statement
   |                      |
   |                      creates a temporary which is freed while still in use
13 |
14 |     while let Some(build_result) = stream.next().await {
   |                                    ------ borrow later used here
   |
   = note: consider using a `let` binding to create a longer lived value
```

And `exmaple/attach.rs` also shows off the exact scenario I described initially. We have to assign `containers` to a local otherwise we have a similar lifetime issue as the earlier example.

https://github.com/softprops/shiplift/blob/5af822c5691a772cad36bd9a69e94419b03d4511/examples/attach.rs#L5-L33

Again, same issue

```
error[E0716]: temporary value dropped while borrowed
  --> src/main.rs:12:27
   |
12 |     let tty_multiplexer = docker.containers().get(&id).attach().await?;
   |                           ^^^^^^^^^^^^^^^^^^^                         - temporary value is freed at the end of this statement
   |                           |
   |                           creates a temporary which is freed while still in use
...
24 | }
   | - borrow might be used here, when `tty_multiplexer` is dropped and runs the destructor for type `Multiplexer<'_>`
   |
   = note: consider using a `let` binding to create a longer lived value
```

[...]

PR #272, opened 2021-02-17
Fix lifetimes

elihunter173:
## What did you implement:

After a dance with borrowck, I updated the lifetimes of problematic methods in `lib.rs` where the return's lifetime was unnecessarily tied to some of the inputs. In particular, `{Containers, Images, Networks, Services, Volumes}::get()` now return a `{Container, Image, Network, Service, Volume}` which is tied to the lifetime of the `Docker` they container rather than the lifetime of the `{Containers, ...}` which created them. This means we no longer have to create unnecessary locals such as `images` in the following code to avoid the compiler complaining about `temporary value dropped while borrowed`
```rust
let images = docker.images();
let mut stream = images.build(&options);
```
Instead, we can write the following as you would intuitively expect to work.
```rust
let mut stream = docker.images().build(&options);
```

Various other methods were also fixed. Generally, anything that returned a stream is now exclusively tied to the lifetime of the `Docker` used to create them (or `Transport` for `transport.rs`) rather than the lifetime of both the `Docker` and whatever they use as input (e.g. `ExecOptions` for `Container::exec()`).

Further, I renamed lifetimes from `'a` to `'docker` and made some lifetimes explicit to make the code easier to read and understand.

Closes: #271 

Ran unit tests.

This should be backwards compatible because in every case I made the lifetimes less restrictive. Nevertheless, I updated the changelog to mention it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 * support for uploading tar to container [#239](https://github.com/softprops/shiplift/pull/239)
 * fix registry authentication to use URL-safe base64 encoding [#245](https://github.com/softprops/shiplift/pull/245)
 * add StopSignal and StopTimeout to ContainerOptionsBuilder [#248](https://github.com/softprops/shiplift/pull/248)
+* update lifetimes of various methods to avoid `temporary value dropped while borrowed` errors [#272](https://github.com/softprops/shiplift/pull/272)
 
 # 0.6.0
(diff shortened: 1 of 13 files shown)


thomaseizinger:
Thank you!

This also fixes a bug where it was impossible to return the `Stream` returned from `logs()` due to lifetime issues!

PR #273, opened 2021-02-22
Fix: Dot should be in quotes

matthiasbeyer:


matthiasbeyer:
I guess this was open long enough and is trivial enough that I merge without a review from a third party :smile:

PR #274, opened 2021-02-26
Update README.md

rahulunair:
Adding tokio as a dependency with all features, this is needed for compiling and running the examples.

## What did you implement:

Closes: #xxx

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,6 +11,7 @@ Add the following to your `Cargo.toml` file
 ```toml
 [dependencies]
 shiplift = "0.7"
+tokio = { version = "1.2.0", features = ["full"] }
 ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -11,6 +11,7 @@ Add the following to your `Cargo.toml` file
 ```toml
 [dependencies]
 shiplift = "0.7"
+tokio = { version = "1.2.0", features = ["full"] }
```
matthiasbeyer, line 14: shiplift does not depend on `1.2.0` of tokio, but `1.0`. The example should reflect that, IMO.

matthiasbeyer (changes requested):

Issue #275, opened 2021-03-05
Why does `Containers::get()` not return Option?


matthiasbeyer:
Hi,

I don't understand why `Containers::get()` does not return an `Option<Container>`. From what I understand, `get()`ing a container by name (which seems to be the argument here) would either return an instance of `Container` or `None` if the named container does not exist.

But the interface does not return an `Option` but simply an instance of `Container`.

I don't understand why.
Maybe the docs are missing something here, maybe I just don't understand.

Can someone please elaborate or, if applicable, fix the documentation/interface?

elihunter173:
Currently `Containers::get()` directly passes the name you provide to `Container::new()` as the ID, so it doesn't do any validation on the name to ensure the container exists, just returns a potentially invalid handles. The docs are definitely lack-luster in this regard and should be improved.

I think it would be a good idea to make it so `Containers::get()` would actually convert the name you provide to an ID and return `Some(Container)` if it is successful or `None` if it cannot find the container.

matthiasbeyer:
I consider this solved with the two PRs merged.

PR #277, opened 2021-03-06
Add Containers::get_checked() for getting an existing container

matthiasbeyer:
## What did you implement:

Containers::get() does always return a handle to a container, which might be invalid.
This function checks whether the container exists before returning either a Some(Container) or None.

Part of: #275

New function `Containers::get_checked()`

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -645,6 +645,25 @@ impl<'docker> Containers<'docker> {
         Container::new(self.docker, name)
     }
 
+    /// Same as `Container::get()`, but checks whether the container exists and returns None if it
+    /// doesn't
+    pub async fn get_checked<S>(
+        &self,
+        name: S,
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -645,6 +645,23 @@ impl<'docker> Containers<'docker> {
         Container::new(self.docker, name)
     }
 
+    pub async fn get_checked<S>(
```
elihunter173, line 648:
```suggestion
    /// Returns a reference to a Container with the given name, if it exists.
    pub async fn get_checked<S>(
```

elihunter173 (commented):
Code looks good but a doc comment would be good

elihunter173 (approved):

matthiasbeyer:
Squashed the fixups, will merge asap.

PR #280, opened 2021-04-01
Allow multiple filters of same type and fix ExitCode filter name

jpommerening:
Hi! :wave:

## What did you implement:

Fixed a problem with `ContainerListOptionsBuilder`, where it was not possible to use multiple filters of the same type.

For example, the following only matches agains the `second` label; the `first` label is dropped:

```rust
// expected to behave like --filter "label=first=value1,second=value2"
// actually behaves like --filter "label=second=value2"
let opts = ContainerListOptions::builder()
    .filter(vec![
        Label("first".to_string(), "value1".to_string()),
        Label("second".to_string(), "value2".to_string()),
    ])
    .build();

```

While at it, I noticed that `ExitCode(_)` is serialized to `exit` whereas the [documentation](https://docs.docker.com/engine/api/v1.41/#operation/ContainerList) expects `exited`.

To verify the serialization of the labels and filters, I added some tests and ran the expected results against my `docker.sock`:

```console
# {"exited":["0"]}
$ curl --unix-socket /var/run/docker.sock 'http:/v1.41/containers/json?all=true&filters=%7B%22exited%22%3A%5B%220%22%5D%7D'
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100     3    0     3    0     0   1000      0 --:--:-- --:--:-- --:--:--  1000
[]
# {"exit":["0"]}
$ curl --unix-socket /var/run/docker.sock 'http:/v1.41/containers/json?all=true&filters=%7B%22exit%22%3A%5B%220%22%5D%7D' | jq
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100    36    0    36    0     0  12000      0 --:--:-- --:--:-- --:--:-- 12000
{
  "message": "Invalid filter 'exit'"
}
# {"label":["label1=value","label2"]}
curl --unix-socket /var/run/docker.sock 'http:/v1.41/containers/json?all=true&filters=%7B%22label%22%3A%5B%22label1%3Dvalue%22%2C%22label2%22%5D%7D' | jq
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100     3    0     3    0     0   1000      0 --:--:-- --:--:-- --:--:--  1000
[]
```

It is now possible to combine multiple filters of the same type, when listing containers. The `ExitCode` filter is now correctly serialized.

### DIFF
--- a/src/container.rs
+++ b/src/container.rs
@@ -513,7 +513,7 @@ impl ContainerListOptionsBuilder {
                 ContainerFilter::Label(n, v) => ("label", format!("{}={}", n, v)),
             };
 
-            param.entry(key).or_insert(Vec::new()).push(value);
+            param.entry(key).or_insert_with(Vec::new).push(value);
         }
         // structure is a a json encoded object mapping string keys to a list
         // of string values
--- a/src/container.rs
+++ b/src/container.rs
@@ -504,14 +504,19 @@ impl ContainerListOptionsBuilder {
         &mut self,
         filters: Vec<ContainerFilter>,
     ) -> &mut Self {
-        let mut param = HashMap::new();
+        let mut param: HashMap<&str, Vec<String>> = HashMap::new();
         for f in filters {
-            match f {
-                ContainerFilter::ExitCode(c) => param.insert("exit", vec![c.to_string()]),
-                ContainerFilter::Status(s) => param.insert("status", vec![s]),
-                ContainerFilter::LabelName(n) => param.insert("label", vec![n]),
-                ContainerFilter::Label(n, v) => param.insert("label", vec![format!("{}={}", n, v)]),
+            let (key, value) = match f {
+                ContainerFilter::ExitCode(c) => ("exited", c.to_string()),
+                ContainerFilter::Status(s) => ("status", s),
+                ContainerFilter::LabelName(n) => ("label", n),
+                ContainerFilter::Label(n, v) => ("label", format!("{}={}", n, v)),
             };
+            if let Some(values) = param.get_mut(key) {
+                values.push(value);
+            } else {
+                param.insert(key, vec![value]);
+            }
         }
         // structure is a a json encoded object mapping string keys to a list
         // of string values
@@ -1409,6 +1414,7 @@ pub struct Exit {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use crate::container::ContainerFilter::{ExitCode, Label, LabelName, Status};
 
     #[test]
     fn container_options_simple() {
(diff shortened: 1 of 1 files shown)


jpommerening:
The `lint` check is failing on files I did not modify…
I'm not sure where to go from here. Is there a Git commit I can rebase my changes onto?

jpommerening:
Rebased for clippy fixes 📎

src/container.rs
```diff
@@ -504,14 +504,19 @@ impl ContainerListOptionsBuilder {
...
+                ContainerFilter::ExitCode(c) => ("exited", c.to_string()),
+                ContainerFilter::Status(s) => ("status", s),
+                ContainerFilter::LabelName(n) => ("label", n),
+                ContainerFilter::Label(n, v) => ("label", format!("{}={}", n, v)),
             };
+            if let Some(values) = param.get_mut(key) {
+                values.push(value);
+            } else {
+                param.insert(key, vec![value]);
+            }
```
matthiasbeyer, line 519: This could be refactored to something like 

```
param.entry(key).and_modify(|values| values.push(value)).or_insert(key, vec![value]);
```

I guess? Can you try and if yes, push it as additional commit (no squash please, so the history reflects the development properly).

matthiasbeyer (changes requested):
Awesome changeset!

If that one comment is solved I'll merge.

Sorry that it took so long for me to react!

jpommerening:
@matthiasbeyer hey thanks!

I'll look at that today! :+1:

src/container.rs
jpommerening: Nice!

jpommerening (commented):
Ohhh … I didn't think of this:
```
error[E0382]: use of moved value: `value`
   --> src/container.rs:518:33
    |
509 |             let (key, value) = match f {
    |                       ----- move occurs because `value` has type `std::string::String`, which does not implement the `Copy` trait
...
517 |                 .and_modify(|values| values.push(value))
    |                             --------             ----- variable moved due to use in closure
    |                             |
    |                             value moved into closure here
518 |                 .or_insert(vec![value]);
    |                                 ^^^^^ value used here after move
```

Can't move the value twice. Do you have an idea what I could do, other than cloning it?
Otherwise I'd stick with the `if/else` from before

matthiasbeyer:
Ah, yes.

Try this:

```bash
git fetch https://github.com/matthiasbeyer/shiplift pr-280
git cherry-pick FETCH_HEAD
```

To apply my idea to clean this up! :smile:

jpommerening:
Ah! :bulb:

I built on that and used `.or_insert_with(Vec::new)` instead of `.or_insert(Vec::new())` to avoid constructing more Vecs than we need :)

src/container.rs
```diff
@@ -513,7 +513,7 @@ impl ContainerListOptionsBuilder {
                 ContainerFilter::Label(n, v) => ("label", format!("{}={}", n, v)),
             };
 
-            param.entry(key).or_insert(Vec::new()).push(value);
+            param.entry(key).or_insert_with(Vec::new).push(value);
```
matthiasbeyer, line 516: Cool! :tada: Today I learned something because of you! :smile:

matthiasbeyer (approved):
I'd merge this now, if you agree! :smile:

src/container.rs
jpommerening: Likewise! 🤝

jpommerening:
Sure, go ahead!
You're on a roll today! 🥇 😄

Issue #288, opened 2021-04-18
Incomplete error message returned by `build`


cswinter:
- Crate version: `0.7.0`
- OS: `Linux 5.4.0-70-generic #78~18.04.1-Ubuntu SMP`
- Output of running `docker version` on the command line:
```
Client: Docker Engine - Community
 Version:           20.10.6
 API version:       1.41
 Go version:        go1.13.15
 Git commit:        370c289
 Built:             Fri Apr  9 22:46:01 2021
 OS/Arch:           linux/amd64
 Context:           default
 Experimental:      true

Server: Docker Engine - Community
 Engine:
  Version:          20.10.6
  API version:      1.41 (minimum version 1.12)
  Go version:       go1.13.15
  Git commit:       8728dd2
  Built:            Fri Apr  9 22:44:13 2021
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.4.4
  GitCommit:        05f951a3781f4f2c1911b05e61c160e9c30eaa8e
 runc:
  Version:          1.0.0-rc93
  GitCommit:        12644e614e25b05da6fd08a38ffa0cfe1903fdec
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0
```

When passing a path to a file rather than a directory as the argument to [examples/imagebuild.rs](https://github.com/softprops/shiplift/blob/master/examples/imagebuild.rs), the only error output is `Docker Error: 500 Internal Server Error`. If at all possible, it would be very useful to report a more specific error message. E.g. when running docker directly with the same args, I get `unable to prepare context: context must be a directory: <path>`.

PR #292, opened 2021-04-21
Display error message for `Error::Fault`

cswinter:
## What did you implement:

Small change to the `Display` implementation of `Error` that includes the `message` returned by docker.
IMO this should be part of the display string since it makes errors much more informative.
E.g., instead of `Docker Error: 404 Not Found` an error would now be displayed as `Docker Error: 404 Not Found: No such container: containername`

Closes: https://github.com/softprops/shiplift/issues/288

manually tested it on one API request 😄

include additional information in error messages

### DIFF
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -77,7 +77,7 @@ impl fmt::Display for Error {
             Error::InvalidResponse(ref cause) => {
                 write!(f, "Response doesn't have the expected format: {}", cause)
             }
-            Error::Fault { code, .. } => write!(f, "{}", code),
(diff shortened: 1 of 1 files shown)


cswinter:
Arguably errors would actually be slightly better by displaying _just_ the message and not the error code, which might not be of interest to users and doesn't convey much additional information. But not sure if docker consistently returns good error messages, maybe sometimes it's good to have the status code as well.

cswinter:
Another option would be to display just the message, except when it's an empty string in which case it shows the error code.

Issue #293, opened 2021-04-27
network inspect tries to parse response as `NetworkInfo`, should be `NetworkDetails`


cswinter:
Calling the network `inspect` function returns an error like `missing field `rx_dropped` at line 1 column 1032`. This is because the `inspect` function returns the `NetworkInfo` type when it should be returning `NetworkDetails`. PR incoming.

PR #294, opened 2021-04-27
Fix network inspect type

cswinter:
## What did you implement:

Fixes return type of network `inspect` function.

Closes: https://github.com/softprops/shiplift/issues/293

Tested manually, before change `inspect` returns parse error, after change, it returns valid network info.

- Fix network `inspect` function


cswinter:
@softprops I think you need to manually kick off build for this PR: https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks

Issue #295, opened 2021-05-20
container logs does not work if tty is enabled


cswinter:
When querying logs for containers with a tty, the logs stream is not multiplexed via headers but just returns the raw output: https://github.com/docker/cli/blob/86e1f04b5f115fb0b4bbd51e0e4a68233072d24b/vendor/github.com/docker/docker/client/container_logs.go#L19
The current logs implementation will parse random STDOUT data as a length header and then tries to perform a `read_exact` with usually an unreasonably large length which runs into EOF error. The observed behavior is that `logs` immediately exits without returning any output.

cswinter:
Identical issue encountered by docker-py: https://github.com/docker/docker-py/issues/630

PR #296, opened 2021-05-20
[RFC] Logs tty

cswinter:
PoC for fixing https://github.com/softprops/shiplift/issues/295. Works well enough as a workaround for my use case, but the current implementation still has some issues which I would like to get feedback on how to best resolve:

- The library user needs to correctly choose between `logs_tty` and `logs`, and choosing the wrong one will either produce additional garbage in stout from multiplexing headers or fail to return any input at all. The approach taken by the docker CLI is to run an `inspect` first to determine if the container uses tty: https://github.com/docker/cli/blob/86e1f04b5f115fb0b4bbd51e0e4a68233072d24b/cli/command/container/logs.go#L53. This would require an additional request, but imo worth it to allows to have a single method without any additional arguments that always returns the correct result.
- The `impl` types returned by the different streams are not compatible, an alternative would be to return `Box<dyn Stream<Item = _> + Unpin>`. Overhead is probably pretty minimal compared to network IO.
- Current implementation does a lot of unnecessary allocations which are probably easily avoidable for someone who knows more about `Stream`-munging than I do.

### DIFF
--- a/src/container.rs
+++ b/src/container.rs
@@ -101,6 +101,20 @@ impl<'docker> Container<'docker> {
         Box::pin(tty::decode(stream))
     }
 
+    pub fn logs_tty(
+        &self,
+        opts: &LogsOptions,
+    ) -> impl Stream<Item = Result<tty::TtyChunk>> + Unpin + 'docker {
+        let mut path = vec![format!("/containers/{}/logs", self.id)];
+        if let Some(query) = opts.serialize() {
+            path.push(query)
+        }
+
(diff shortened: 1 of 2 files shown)


cswinter:
Oh one more issue, the `logs_tty` function now never terminates as long as the container is alive even when you don't set `tail`. Not sure how to correctly detect the end of the current output.

matthiasbeyer:
Requested @softprops review here because I really don't know :shrug:

PR #298, opened 2021-06-07
Added volume driver support

moritzheiber:
## What did you implement:

Added `Driver` and `DriverOpts` support [according to the Docker API](https://docs.docker.com/engine/api/v1.41/#operation/VolumeCreate).

## How did you verify your change

Two tests were added to verify the proper serialization of the added options.


matthiasbeyer (approved):
Cool! Especially that you added tests!

IMO this looks good, will merge if CI allows me to! :laughing:

Issue #301, opened 2021-06-16
We have too many unwrap() calls in the codebase


matthiasbeyer:
Doing a quick analysis:

```bash
grep -rni unwrap -c src/ | cut -d: -f 2 | awk '{sum+=$1} END {print sum}'
```

gives me a shocking 53. Of course some of those (probably the majority) is in tests, where it is (IMO) appropriate to unwrap right away, but a lot of them are not in tests (after a quick `less` on some files).

This is rather easy to tackle IMO, maybe even for a beginner (one unwrap fixed at a time in one PR even, I am happy to mentor!)

---

I'll mark this as good first issue and submit it to this-week-in-rust!

Sirneij:
I have give it a try.

matthiasbeyer:
Cool. If you need any help, feel free to ping me on matrix @musicmatze:beyermatthi.as or via mail (see my github profile), or just open a draft PR and ask there, whatever works for you!

p14c31355:
@matthiasbeyer Hi! I am @YoshitakaNaraoka , Japanese Rustacean egg.
This PR is abandoned for a long time, please assign me.

PR #308, opened 2021-07-15
Replace some of the unwrap() calls

Vinegret43:
## What did you implement:
Replace some of "unwrap()" calls either with a question mark "?" or a match statement (Had to change return type of some functions (Didn't do anything to public functions/methods) and add new variants and traits for Error enum (See the changelog section)).
Also fix all clippy warnings.

Notes:
1. function "insert" in container.rs uses ```std::io::Result``` return type instead of built-in Result type in this crate. This is because i would have to implement something like JSON variant for the Error enum, then add different kinds for JSON errors as well, and increase complexity of code. This return type just works, so i used it instead

Closes: #301 (Only partially, there are still many unwrap calls outside of tests and examples)

Building and testing the project was successful

(All changes are backwards-compatible)
Added two new variants for the Error enum: Ssl and Mime. Also implemented traits ```From<openssl::error::ErrorStack>``` and ```From<mime::FromStrError>``` for Error enum

### DIFF
--- a/src/container.rs
+++ b/src/container.rs
@@ -75,7 +75,7 @@ impl<'docker> Container<'docker> {
         psargs: Option<&str>,
     ) -> Result<Top> {
         let mut path = vec![format!("/containers/{}/top", self.id)];
-        if let Some(ref args) = psargs {
+        if let Some(args) = psargs {
             let encoded = form_urlencoded::Serializer::new(String::new())
                 .append_pair("ps_args", args)
                 .finish();
@@ -386,7 +386,7 @@ impl<'docker> Container<'docker> {
             .append_pair("path", &path.to_string_lossy())
             .finish();
 
-        let mime = "application/x-tar".parse::<Mime>().unwrap();
+        let mime = "application/x-tar".parse::<Mime>()?;
 
         self.docker
             .put(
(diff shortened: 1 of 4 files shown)


src/errors.rs
```diff
@@ -77,6 +93,8 @@ impl fmt::Display for Error {
             Error::InvalidResponse(ref cause) => {
                 write!(f, "Response doesn't have the expected format: {}", cause)
             }
+            Error::Ssl(ref err) => err.fmt(f),
+            Error::Mime(ref err) => err.fmt(f),
```
thomaseizinger, line 97: Instead of printing the source error, we could also return it in `StdError::source` and either print nothing here or a message "on top" of what happened.

Also see: https://users.rust-lang.org/t/do-people-not-care-about-printable-error-chains-a-k-a-how-to-nicely-implement-display-for-an-error/35362/2

Vinegret43: Made two new commits. Not really sure if second commit was necessary, please check it and i will undo it if needed. The problem with it is that you said that i could either "print nothing here" (Which is definitely not an option) or print a message "on top" of what happened, which will output something like this with Mime type: ```Mime error: mime parse error: <ParseError>``` (See how mime::FromStrError displays itself at its [source code](https://docs.rs/mime/0.3.16/src/mime/lib.rs.html#82-86)). As you can see it does not look really good, first two statements just repeat each other. I would remove this "Mime error:" at the beginning and leave just ```err.fmt(f)``` in code how it was before. If i'm wrong and i undo these changes, i might need to redo them again, so just wanna see what you think about that before doing anything

thomaseizinger: If we don't have anything more meaningful to print on top of the original error, why is printing nothing not an option? Together with returning the `source`, libraries like `anyhow` will print the entire chain of errors and thus print the "inner" error.

But, I think the better option would be to move away from a generall "Ssl" error and be more specific of what failed. I commented inline with some ideas, feel free to ignore if you find them too verbose :)

src/container.rs
```diff
@@ -386,7 +386,7 @@ impl<'docker> Container<'docker> {
             .append_pair("path", &path.to_string_lossy())
             .finish();
 
-        let mime = "application/x-tar".parse::<Mime>().unwrap();
+        let mime = "application/x-tar".parse::<Mime>()?;
```
thomaseizinger, line 389: Can this actually ever fail? `unwrap`/`expect` seems to be ok here.

src/docker.rs
```diff
@@ -48,24 +48,18 @@ fn get_http_connector() -> HttpConnector {
...
+        connector.set_cipher_list("DEFAULT")?;
         let cert = &format!("{}/cert.pem", certs);
         let key = &format!("{}/key.pem", certs);
-        connector
-            .set_certificate_file(&Path::new(cert), SslFiletype::PEM)
-            .unwrap();
-        connector
-            .set_private_key_file(&Path::new(key), SslFiletype::PEM)
-            .unwrap();
+        connector.set_certificate_file(&Path::new(cert), SslFiletype::PEM)?;
```
thomaseizinger, line 58: Instead of the generic Ssl error, we could map to a variant `Error::InvalidCertificate`. That would give us something meaningful to say in our `Display` impl without having to print the inner error.

Vinegret43: SSL error can be returned not just here, it will be generated, for example in docker.rs at line 54: ```let mut connector = SslConnector::builder(SslMethod::tls())?;```, and on line 59: ```connector.set_private_key_file(&Path::new(key), SslFiletype::PEM)?;``` (And there are more).  So we can't really guarantee that all of these errors will be caused by an invalid certificate, and in my opinion, it's better to make just a general SSL error, containing an inner error

src/container.rs
Vinegret43: Yes, this is right. Thanks for your advice a lot

src/docker.rs
thomaseizinger: Yes, we would need more than one variant, not just `InvalidCertificate` but also `InvalidPrivateKey` etc.

Which is why I said "feel free to ignore if you think it becomes too verbose" :)

A general Ssl error is certainly more concise but the error messages will be less helpful because it is not as clear where the error originated.

